PR #2, opened 2017-05-15
Configures package.json for npm consumption

Ollo:
- Adds basic button partial to test integration

cc/ @saralohr @williamdingwall

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,0 +1,19 @@
+{
+  "name": "teamsnap-ui",
+  "version": "0.1.0",
+  "description": "a CSS component library for TeamSnap",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/teamsnap/teamsnap-ui.git"
+  },
+  "author": "",
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/teamsnap/teamsnap-ui/issues"
+  },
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -0,0 +1,19 @@
...
+  "description": "a CSS component library for TeamSnap",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/teamsnap/teamsnap-ui.git"
+  },
+  "author": "",
```
Ollo: Wasn't sure if there was a `design@teamsnap.com` alias or something that would be useful here.

package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "teamsnap-ui",
+  "version": "0.1.0",
+  "description": "a CSS component library for TeamSnap",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
```
Ollo: This section has some options to write specific action that should take place after pushing to git, installing into a project etc.. <https://docs.npmjs.com/misc/scripts>

src/Button/Button.scss
```diff
@@ -0,0 +1,4 @@
+.Button {
```
Ollo: Just put this in as a point to test with.. directory structure is all up in the air.

mikereinmiller: was just about to comment on that.  Good call, I'm good with this for initial testing.  Nice work.

PR #4, opened 2017-05-16
Add initial structure and button component

saralohr:
This PR adds a directory structure seeded with some of the initial settings for defining components. The variables in `config` are pulled from Classic since all the themes are based on these names. The Button component kicks off a new [BEM naming convention](https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md).

I tried out including the variable `config.scss` in the `.Button` component. Let's see how that works out. We can modify if that doesn't seem right.

Some of the structure is built to facilitate the inclusion of this library as a submodule within the style guide. For example, adding `index`es within each directory and in the top level makes it easy to grab just one file for compilation in the style guide:

button

### DIFF
--- a/src/components/button.scss
+++ b/src/components/button.scss
@@ -70,10 +70,10 @@ $Button-padding: 0 ($Button-height / 1.75);
  * Modifier: Button Primary
  */
 
-$Button--primary-bg: $color-primary;
+$Button--primary-bg: $color-highlight;
 $Button--primary-bg-hover: darken($Button--primary-bg, 5%);
 $Button--primary-color: $ui-white;
-$Button--primary-border-color: darken($Button--primary-bg, 7.5%);
+$Button--primary-border-color: $color-highlight--dark;
 $Button--primary-border-color-hover: darken($Button--primary-bg, 10%);
 
 .Button--primary {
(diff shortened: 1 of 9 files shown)


src/components/button.scss
```diff
@@ -0,0 +1,219 @@
...
+$Button-border-radius: $border-radius-medium;
+$Button-border-width: $border-width-small;
+$Button-border-bottom-width: $border-width-medium;
+$Button-disabled-opacity: 0.4;
+$Button-font-family: $base-font-family;
+$Button-font-size: $base-font-size;
+$Button-font-weight: $med-font-weight;
+$Button-height: 34px;
+$Button-line-height: $Button-height - 2;
+$Button-padding: 0 ($Button-height / 1.75);
```
saralohr: The `height`, `line-height` and `padding`, pulled from the current button, feel a little magic-numbery to me, but I was wanting to match the existing buttons exactly so there is no difference between the two if they are living side-by-side.

mikereinmiller:
I really like how this is looking!  My only thought/concern (and maybe only time will tell), I'm curious if a single file for all the styles related to Base, Modifiers and Descendents could become a little unwieldy?

src/components/button.scss
```diff
@@ -0,0 +1,217 @@
...
+ * Modifier: Button Primary
+ */
+
+$Button--primary-bg: $color-primary;
+$Button--primary-bg-hover: darken($Button--primary-bg, 5%);
+$Button--primary-color: $ui-white;
+$Button--primary-border-color: darken($Button--primary-bg, 7.5%);
+$Button--primary-border-color-hover: darken($Button--primary-bg, 10%);
+
+.Button--primary {
```
williamdingwall: Looking Good!  I like the positive / negative naming!

Can we cut out button--primary from here.  I know there are a few stragglers, but that was one we started to cut during the reskin.  It's in classic right now, because it's still in use in a few spots, but will be phased out.

buttons

saralohr: Sure, I'll remove the blue button. I wonder if it makes sense to name the `--positive` version `--primary` instead? The positive/negative distinction works really well with the green/red, and as a counterpoint to another primary button, but if there is no other primary button then it might make sense to swap it? I'm cool either way, just curious on your thoughts @williamdingwall.

saralohr:
> My only thought/concern (and maybe only time will tell), I'm curious if a single file for all the styles related to Base, Modifiers and Descendents could become a little unwieldy?

Yeah, I could see multiple files becoming unwieldy too... but I'm not opposed to trying it. Let's see how it's working out with everything in one file and if it is too heavy we can split it out.

src/components/button.scss
williamdingwall: @saralohr that makes sense, I like switching that CTA button to "primary"

mikereinmiller:
@saralohr That's a good point too, I suppose either one could become a bit of a beast.

src/components/button.scss
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  &.is-active {
+    background-color: $Button-bg-hover;
+    box-shadow: 0 1px 0 $Button-border-color inset;
+  }
+
+  &.is-disabled,
+  &:disabled {
+    &,
```
Ollo: I haven't seen this before what does it do?

saralohr: This is grabbing the base `is-disabled` state in addition to `:hover` and `:active` so that they are all the same.

Ollo: neat! I thought you just needed to nest them to accomplish that. Does the comma make it more precise?

saralohr: Oh, not sure, actually!! I'll have to experiment and see. To be honest, I pulled this from the previous button styles :)

PR #7, opened 2017-05-18
Adjust for pre-existing styles

saralohr:
This PR adjusts for the fact that we are loading global styles currently in LFE and NJW, so we don't need to redefine those base element styles here. Also adjusts the `Button--primary` component to account for `color` being applied to the base `button:hover` element in Classic.


mikereinmiller:
It might be nice to have some of the base styles loaded in this ui (if a component depends on it).  @Ollo Would these be nice to pull down locally or does it not really matter if they are in the global space?  I'm honestly not sure I guess.

saralohr:
> It might be nice to have some of the base styles loaded in this ui

This is what I was thinking too... and starting doing so for the sake of the patterns toolkit, but then I realized that since components are living on top of the Classic styles in production, we should probably be replicating that environment for any local work. So I actually just pulled a link to the `app-team.css` in the patterns library for now. Not sure how stuff works locally with React, but there are some opinionated styles in classic on basic elements like `button`, `input` and `table` that I wouldn't ever replicate for something new... so essentially we would need to pull _all_ the classic styles into our nice clean CSS components repo to recreate the prod scenario. In the future, if this CSS is ever living somewhere _without_ the Classic styles, we would need to create new base typography and reset styles in this repo.

Ollo:
I think it would be super helpful to have the classic styles loaded. Right now when working on react components in LFE you have to just _hope_ they look correct during local development and test by loading through classic due to the additional styles it provides.

saralohr:
> I think it would be super helpful to have the classic styles loaded. Right now when working on react components in LFE you have to just hope they look correct during local development and test by loading through classic due to the additional styles it provides.

Any ideas on how to achieve this?

PR #8, opened 2017-05-24
first pass on StepNav

williamdingwall:
Adding Step Nav component from registration and updating to make more easily composable..

screen shot 2017-05-24 at 10 17 21 am

screen shot 2017-05-24 at 10 17 31 am

for reference : https://github.com/teamsnap/ui-patterns/pull/1

### DIFF
--- a/src/base/config.scss
+++ b/src/base/config.scss
@@ -114,3 +114,61 @@ $border-width-medium:  2px;
  $base-font-weight:  normal !default;
  $med-font-weight:   600 !default;
  $bold-font-weight:  bolder !default;
+
+  // Heading Font Settings
+
+  // TODO : take audit of the museo fonts used and clean up heading font variables
+  $museo-100:   'MuseoSansRounded100Regular', 'Open Sans', Arial, sans-serif !default;
+  $museo-300:   'MuseoSansRounded300Regular', 'Open Sans', Arial, sans-serif !default;
+  $museo-500:   'MuseoSansRounded500Regular', 'Open Sans', Arial, sans-serif !default;
+  $museo-700:   'MuseoSansRounded700Regular', 'Open Sans', Arial, sans-serif !default;
+  $museo-900:   'MuseoSansRounded900Regular', 'Open Sans', Arial, sans-serif !default;
+  $museo-1000: 'MuseoSansRounded1000Regular', 'Open Sans', Arial, sans-serif !default;
+
+  $heading-font-family:                                  $museo-700 !default;
+  $sub-heading-font-family:                              $open-sans !default;
+
+  $heading-font-color:                               $color-primary !default;
+  $sub-heading-font-color:                         $color-secondary !default;
+
+  // TODO : Revisit typographic scale to find common ground with marketing
+  $heading-size--primary:             round($base-font-size * 2.25) !default;
+  $heading-size--secondary:           round($base-font-size * 1.75) !default;
+  $heading-size--minor:               round($base-font-size * 1.25) !default;
+  $heading-size--small:               round($base-font-size * 1.00) !default;
+  $heading-size--xs:                  round($base-font-size * 0.85) !default;
+
+  $heading-line-height--primary:                                  2 !default;
+  $heading-line-height--secondary:                              1.7 !default;
+  $heading-line-height--minor:                                  1.5 !default;
+  $heading-line-height--small:                                  1.4 !default;
+  $heading-line-height--xs:                                    1.15 !default;
+
+/**
+ * Spacing
+ */
+
+$block-spacing:                           25px !default;
+$block-spacing--xsmall:                    7px !default;
+$block-spacing--small:                    15px !default;
+$block-spacing--large:                    40px !default;
+
+$element-spacing:                          9px !default;
+$element-spacing--small:                   6px !default;
+$element-spacing--large:                  15px !default;
+
+$horizontal-block-spacing:                25px !default;
+$horizontal-block-spacing--small:         15px !default;
+$horizontal-block-spacing--large:         40px !default;
+
+$horizontal-element-spacing:               9px !default;
+$horizontal-element-spacing--small:        6px !default;
+$horizontal-element-spacing--large:       14px !default;
+
+$vertical-block-spacing:                  25px !default;
+$vertical-block-spacing--small:           15px !default;
+$vertical-block-spacing--large:           40px !default;
+
+$vertical-element-spacing:                 7px !default;
+$vertical-element-spacing--small:          4px !default;
+$vertical-element-spacing--large:         12px !default;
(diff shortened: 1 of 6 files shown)


src/base/config.scss
```diff
@@ -114,3 +114,61 @@ $border-width-medium:  2px;
...
+
+  $heading-line-height--primary:                                  2 !default;
+  $heading-line-height--secondary:                              1.7 !default;
+  $heading-line-height--minor:                                  1.5 !default;
+  $heading-line-height--small:                                  1.4 !default;
+  $heading-line-height--xs:                                    1.15 !default;
+
+/**
+ * Spacing
+ */
```
saralohr: Should we add a TODO to simplify and refine spacing variables?

src/components/step-nav.scss
```diff
@@ -0,0 +1,197 @@
...
+$StepNav-border--small: 2px;
+
+// * 2. Define Default StepNav
+
+.StepNav {
+  padding-left: ($StepNav-size * 1.25);
+  position: relative;
+  @include clear();
+  padding-bottom: $vertical-block-spacing--small;
+  &.steps-2 {
```
saralohr: These `steps-*` feel like `.StepNav` modifiers to me. The numbers are a little tricky with naming since they don't have a case, but these could be something like `StepNav--4Steps` or `.StepNav--fourSteps`... and wouldn't need to be nested within the `StepNav` because the naming convention connects them as a modifier.

src/components/step-nav.scss
```diff
@@ -0,0 +1,197 @@
...
+    }
+  }
+  &.steps-4 {
+    .StepNav-Step {
+      width: 33%;
+    }
+  }
+}
+
+.StepNav-Step {
```
saralohr: [Children](https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md#ComponentName-descendentName) and modifiers use camel case, so it would be `.StepNav-step`, `.StepNav-item`.

src/components/step-nav.scss
```diff
@@ -0,0 +1,197 @@
...
+      transition: border-color .35s ease-in-out;
+    }
+    &:before {
+      background-color: $color-primary;
+      transition: background-color .35s ease-in-out;
+    }
+  }
+
+  &.is-active,
+  &.is-enabled a:hover {
```
saralohr: Will this `a` always be a `StepNav-step`? If so we should probably use the class selector rather than the element selector.

src/components/step-nav.scss
```diff
@@ -0,0 +1,197 @@
...
+  }
+}
+
+// * 3. Modifier : Small StepNav (coming soon)
+
+// * 4. Modifier : Titled StepNav (coming soon)
+
+// * 5. Responsive Mixin and Usage
+
+@mixin Responsive-StepNav($size, $border) {
```
saralohr: Not sure what's going on with this guy? I don't see it being used, and could this just be a modifier of `StepNav` ie `StepNav--responsive`?

src/base/config.scss
williamdingwall: sure, adding that in now.

src/components/step-nav.scss
williamdingwall: sounds good

src/components/step-nav.scss
williamdingwall: not currently, it only becomes an `a` when enabled, so the future steps are just spans.

src/components/step-nav.scss
williamdingwall: Moved away from that because the step count really should be defined, whereas a modifier seems more optional... I'm fine with that approach though.

src/components/step-nav.scss
williamdingwall: I doubt it will be used in the app, but could see it being used in other places outside the app and it was already defined from registration. Could make it ... Actually I think I might use it to define the small modifier, just allows you to resize the step nav : https://github.com/teamsnap/teamsnap/blob/master/app/assets/stylesheets/new-skin/registration/_navigation.scss#L145

I'll just put up another commit adding the small modifier and update that to see what you think

src/components/step-nav.scss
saralohr: Hmm, interesting. Not sure how this is being implemented and updated, but could it get a modifier if it is a link? ie `StepNav-step--link`... ?

Ollo:
> I also wonder if you explored whether this component might be more easily achieved with a flexbox layout? Might remove the need for things like the --steps* modifiers... there might be reasons why this doesn't work, but something to consider.

I'm also curious on this  idea, as when it we implement with js I would likely want to set the steps dynamically so I can also track state for which one should also have the active class applied. If we went with flexbox I believe it could dynamically space the icons so we don't have to rely on managing classes to size them. We could find a solution using classes too just something I was curious about. Great work so far!  🏄

src/base/config.scss
williamdingwall: I added, and moved to the top with the rest of the TODO's.

williamdingwall:
Thanks for the feedback! Thanks!  added in the modifiers for this too.. had to make some changes with that titled element addition.

small
screen shot 2017-05-24 at 2 53 35 pm

titled
screen shot 2017-05-24 at 2 52 43 pm

src/components/step-nav.scss
```diff
@@ -0,0 +1,234 @@
...
+    top: ($StepNav-size / 2);
+    right: ($StepNav-size * 1.25);
+  }
+
+  &:first-child {
+    position: absolute;
+    left: 0;
+    top: 0;
+    .StepNav-stepLink,
+    .StepNav-stepItem {
```
saralohr: It looks like there is also a `StepNav-link` used for the hover property: https://github.com/teamsnap/teamsnap-ui/pull/8/files#diff-17dbf635491c3ba68c49aedc3b3af731R84

I'm wondering if this link vs item could be simplified. Can the presence of this different style be determined based on the state of the parent? ie via `is-enabled` or `is-active`?  Or could it _just_ be determined by the presence of a different class, ie `StepNav-stepLink`?

src/components/step-nav.scss
```diff
@@ -0,0 +1,234 @@
...
+
+$StepNav-size--small: 40px;
+$StepNav-border--small: 2px;
+
+// * 2. Define Default StepNav
+
+.StepNav-steps {
+  padding-left: ($StepNav-size * 1.25);
+  position: relative;
+  @include clear();
```
williamdingwall: need the position relative, but not the clear. nice catch

src/components/step-nav.scss
williamdingwall: yeah, that could just be stepItem and the enabled state could handle the rest - good call

src/components/step-nav.scss
```diff
@@ -0,0 +1,234 @@
...
+
+$StepNav-size: 60px;
+$StepNav-border: 4px;
+
+$StepNav-size--small: 40px;
+$StepNav-border--small: 2px;
+
+// * 2. Define Default StepNav
+
+.StepNav-steps {
```
williamdingwall: sounds good

williamdingwall:
Thanks!  Just put in those most recent changes

PR #10, opened 2017-05-26
Npm package updates

### LINKED ISSUE #3: expose versioning for npm installations
As an application author I would like to load a specific version of the teamsnap-ui via npm. 

```
npm install --save teamsnap-ui@0.2.0 
```

I think this requires us to configure an organization on NPM.

Ollo:
- Configures and publishes to npm under teamsnap org 🎉  https://www.npmjs.com/package/@teamsnap/teamsnap-ui 
- Configures `sass-lint` and starter rules config 
- Configures npm scripts for testing / publishing updates fixes #3

### DIFF
--- a/sass-lint.yml
+++ b/sass-lint.yml
@@ -1,0 +1,55 @@
+#########################
+## Starter config file
+#########################
+# Linter Options
+options:
+  # Don't merge default rules
+  merge-default-rules: false
+  # Set the formatter to 'html'
+  formatter: html
+  # Output file instead of logging results
+  output-file: 'linters/sass-lint.html'
+  # Raise an error if more than 50 warnings are generated
+  max-warnings: 50
+# File Options
+files:
+  include: 'sass/**/*.s+(a|c)ss'
+  ignore:
+    - 'sass/vendor/**/*.*'
+# Rule Configuration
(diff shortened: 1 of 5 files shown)


sass-lint.yml
```diff
@@ -0,0 +1,55 @@
+#########################
+## Starter config file
+#########################
+# Linter Options
```
saralohr: I just checked out the branch and ran the linter. Getting lots of warnings for things like `property-sort-order` and `class-name-format`. Not a big deal -- but maybe we can add a TODO to customize it a bit more to fit our style?

Is it doing what you need as far as calling out errors that will impact the build to React?

Ollo: I believe we would need to coordinate some additional items with devops but this story was more to get it formally into npm (versioning etc) and to just get a linter in to start customizing rules. Does this impact current workflow for you all?  Definitely feel free to update the rules as desired.

saralohr: Gotcha 👍 It doesn't impact our workflow since we're not running the linter as part of our build currently. I'd like to incorporate it in the future though.

Nice job on the NPM stuff. It's awesome seeing it live!

PR #12, opened 2017-05-30
Add flex, grid and size

saralohr:
+ Updates comment style so comments aren't output in CSS
+ Make spacing more consistent in `config.scss`
+ Adds `Grid` component, `size` utilities, and `flex` utilities borrowed from SUITCSS with a few exclusions, and Sass-ified.

Documented in `ui-patterns`: https://github.com/teamsnap/ui-patterns/pull/7

### DIFF
--- a/src/base/config.scss
+++ b/src/base/config.scss
@@ -3,11 +3,10 @@
 // TODO Revisit typographic scale to find common ground with marketing
 // TODO take audit of the museo fonts used and clean up heading font variables
 
-/**
- * State Color Settings
- *
- * These should stay consistent across themes
- */
+
+// * State Color Settings
+// *
+// * These should stay consistent across themes
 
 $color-success:        #7fc243 !default;
 $color-success--light: lighten($color-success, 45%) !default;
@@ -28,11 +27,10 @@ $color-info:           #494949 !default;
 $color-info--light:    #797979 !default;
 $color-info--dark:     #292929 !default;
 
-/**
- * Brand Colors
- *
- * These can change for different themes
- */
+
+// * Brand Colors
+// *
+// * These can change for different themes
 
 $color-primary:          #1667AC !default;
 $color-primary--light:   #0279CE !default;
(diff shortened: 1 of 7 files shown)


src/base/config.scss
```diff
@@ -3,11 +3,10 @@
 // TODO Revisit typographic scale to find common ground with marketing
 // TODO take audit of the museo fonts used and clean up heading font variables
 
-/**
- * State Color Settings
- *
- * These should stay consistent across themes
- */
+
+// * State Color Settings
```
williamdingwall: I like seeing this

src/utils/size.scss
```diff
@@ -0,0 +1,36 @@
+// * Percentage based size classes for a 24 column grid
+
+$cols: 24;
+
+@for $i from 1 through ($cols) {
```
williamdingwall: I like this approach for defining the widths

saralohr:
> I wonder if it'd be a good idea to start adding responsive utilities too. Just turned off now.

I was wondering this as well. By turned off do you mean commented out? 

Sizing is probably the main one that we would need responsive utilities for, and everything else we could add on an as-necessary basis.

williamdingwall:
Yeah I would mostly agree with that. For turning off, I was thinking something more along these lines...
```
$responsive: false !default;

$screen-xs:  480px !default;
$screen-sm:  768px !default;
$screen-md:  992px !default;
$screen-lg: 1200px !default;

$breakpoints: (
  'xs'      'screen and (min-width: #{$screen-xs})',
  'sm'     'screen and (min-width: #{$screen-sm})',
  'md'     'screen and (min-width: #{$screen-md})',
  'lg'	      'screen and (min-width: #{$screen-lg})'
);

/* Loop through widths .u-size#{$i}of#{$cols} to create base widths */

@if $responsive == true {
  /* Loop through breakpoints variable and widths .u-size#{$i}of#{$cols}--#{$alias} to create responsive widths */
}
```

that way if we did want to build in responsive versions of components, or add in container width variables we could base it all off a single $responsive variable that can be turned off or on per project.

saralohr:
I like that idea a lot @williamdingwall. If it's cool with you I'll add that as a new ticket and separate PR.

PR #13, opened 2017-06-07
update button naming for bugfix

mikereinmiller:
Jenkins is throwing an error with my changes to the Button filename in https://github.com/teamsnap/league-frontend/pull/240#issuecomment-306888363

I *think* this is because it is looking for `Button.scss` but the file is named `button.scss`.  Why this works locally in test but bails on the server I'm not for sure.

Do these changes make sense, did miss anything?  Once this is approved I'll try versioning and publishing to NPM.


mikereinmiller:
@Ollo true.  That makes sense!

PR #14, opened 2017-06-08
Add SVG icons

saralohr:
icons


saralohr:
@ollo Not sure what's best for you to include SVG icons in React components. We currently use a helper to make including them in the pattern library a breeze. Let me know if these should be within an `icon` directory under `components` or anything else that might make implementation more lean.

src/components/icon.scss
```diff
@@ -0,0 +1,25 @@
...
+  vertical-align: middle;
+  top: -1px; // 1
+  fill: currentColor; // 2
+}
+
+// * Modifier -- large icon relative to parent text
+
+$Icon--large-size: 1.25em;
+
+.Icon--large {
```
williamdingwall: I keep going back and forth on this, but from a design standpoint have a hard time seeing modifiers being used with icons.  Typically that icon size is going to be set by the component using it (mostly at 90%), but in the cases where it needs a specific size I have a hard time seeing that being standardized in any way... at least right now, the more we move to design scales that could change.

saralohr: I think if we have a font-size increment ratio that we are using, this becomes easier. Then we can have font sizing utilities that increase or decrease font size, and can be used on icons too. It sounds like you are working on some typographic scale stuff, so maybe we wait until then to implement any icon sizing. I know in the past it's been very useful to have a way to size icons differently in different situations, without having to rely on a component for this. We can see what we need once we start actually using them. I'll remove for now.

saralohr:
> I'm gonna makes some updates to the Invision prototype icon section based off these, and run another icon audit of he app. May have updates after that.

Okay, I will merge for now. We can update as necessary.

PR #15, opened 2017-06-08
Add breakpoint variables, responsive sizing and reduced fractions

saralohr:
size

### DIFF
--- a/src/base/config.scss
+++ b/src/base/config.scss
@@ -192,3 +192,15 @@ $vbs--large:             $vertical-block-spacing--large;
 $ves:                         $vertical-element-spacing;
 $ves--small:           $vertical-element-spacing--small;
 $ves--large:           $vertical-element-spacing--large;
+
+
+// * Responsive Breakpooints
+
+$responsive: false !default;
+
+$breakpoints: (
+  'xs':  480px,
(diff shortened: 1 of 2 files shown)


saralohr:
> I think we should double check our supported browser list, run it by some PMs and Dev leads, and make sure were only supporting modern browsers

Sure, will do. Pretty much just need to worry about IE9 and down. [Flexbox support](http://caniuse.com/#feat=flexbox) these days  is excellent.  I did get user data from Charlie that put IE9 users at .07% of our total users. To me this doesn't justify _not_ using a modern feature, since they will still be able to access all the content, it just won't look as nice. But we should double check our browser support policy. There are a couple fallback options for flex, or we could use the older version of SUIT's grid which uses `display: inline-block` instead, and lacks a few of the nicer features of flexbox but is still totally workable. We'd need to drop the flex utilities obviously too.

> also how much device testing have you done with flexbox?

Lots... all my projects for the past year were responsive and used flexbox, so I've done extensive device testing with flexbox. Devices don't have issues as much as some known bugs in IE and older versions of web browsers that there are well documented work-arounds for. [Flexbugs](community-curated list of flexbox issues and cross-browser workarounds for them) is a great resource for this.

PR #16, opened 2017-06-12
Updating the step nav to support longer text

williamdingwall:
before : 
screen shot on 2017-06-06 at 13-48-58

after : 
screen shot 2017-06-12 at 10 32 21 am

### DIFF
--- a/src/components/step-nav.scss
+++ b/src/components/step-nav.scss
@@ -17,13 +17,13 @@
 // * 1. Settings
 
 $StepNav-size: 60px;
-$StepNav-border: 4px;
-$StepNav-paddedWidth: $StepNav-size + ($StepNav-border * 2);
+$StepNav-spacing: 4px;
+$StepNav-paddedWidth: $StepNav-size + ($StepNav-spacing * 2);
 $StepNav-titleWidth: 150px;
 
 $StepNav-size--small: 40px;
-$StepNav-border--small: 2px;
-$StepNav-paddedWidth--small: $StepNav-size--small + ($StepNav-border--small * 2);
+$StepNav-spacing--small: 2px;
+$StepNav-paddedWidth--small: $StepNav-size--small + ($StepNav-spacing--small * 2);
 
 // * 2. Define Default StepNav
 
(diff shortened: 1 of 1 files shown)


src/components/step-nav.scss
```diff
@@ -10,46 +10,48 @@
 // * 5. Titled Modifier
 
 // TODO : document global dependencies
-// TODO : Add Titled Modifier
+// TODO : further responsive cleanup, especually on titled setp nav
```
dmccraw: especially

src/components/step-nav.scss
```diff
@@ -98,21 +100,24 @@ $StepNav-border--small: 2px;
...
   display: block;
   text-align: center;
   margin: 0 auto;
   color: $ui-divider;
-  padding-top: $StepNav-border;
   font-family: $base-font-family;
   font-size: $heading-size--small;
+  position: absolute;
+  width: $StepNav-titleWidth;
+  bottom: -20px;
```
saralohr: I wonder if doing something like `top: 100%` and building some padding into the title might be less fragile in case the text breaks into two lines:
 
screen shot 2017-06-12 at 10 35 34 am

src/components/step-nav.scss
```diff
@@ -10,46 +10,48 @@
...
-// TODO : Add Titled Modifier
+// TODO : further responsive cleanup, especially on titled setp nav
 
 //------------------------------------------------------------//
 
 // * 1. Settings
 
 $StepNav-size: 60px;
 $StepNav-border: 4px;
+$StepNav-paddedWidth: $StepNav-size + ($StepNav-border * 2);
```
saralohr: Nit: This threw me off a little... since `$StepNav-border` is used to define both the border and the padding. In this case, it's referring to the border value as _padding_, correct? I wonder if adding a variable like `$StepNav-padding: $StepNav-border` might make this clearer.

williamdingwall: Yeah that's fair, I updated to $StepNav-spacing to be a little more generic since it should cover both that border width and padding

src/components/step-nav.scss
williamdingwall: good call! updated

PR #18, opened 2017-06-21
Design updates merge

williamdingwall:
Implementing design updates from here : https://github.com/teamsnap/teamsnap/pull/6862

as well as adding in some form components


williamdingwall:
Thanks @mikereinmiller !  Just added a panel component too and a rough demo around building a form with it here : https://github.com/teamsnap/ui-patterns/pull/13

It's still pretty rough for sure and in need of some styling, but the general idea is there

PR #22, opened 2017-07-12
Reorganize SCSS files

a-dg:
This reorganizes the SCSS files in preparation for incorporating component styles from Classic.


Ollo:
We will need to test this update in League Frontend. cc/ @mikereinmiller

src/fonts/index.scss
```diff
@@ -0,0 +1,3 @@
+@import "museo";
+// @import "open-sans";
```
williamdingwall: Not sure why this is commented out - we can include

src/teamsnap-ui.scss
```diff
@@ -1,3 +1,5 @@
+@import "config/index";
 @import "base/index";
+@import "fonts/index";
```
williamdingwall: minor nit, probably doesn't matter functionally, but what do you think about moving fonts above config?

src/utils/size.scss
```diff
@@ -1,4 +1,4 @@
-@import "../base/config.scss";
+@import "../config/config.scss";
```
williamdingwall: We'll need to do this for all the components too..  `@import "../config/config.scss";` needs to be included for each one so they can all be imported and composed upon individually within react.

src/base/index.scss
```diff
@@ -1,4 +1,4 @@
-@import "functions";
-@import "mixins";
-@import "config";
+@import "reset";
 @import "typography";
+@import "lists";
+@import "tables";
```
saralohr: Should lists and tables be included? I don't see files for these.

saralohr:
> @saralohr any thoughts on this structure with the new config and base styles folders?

This seems fine to me.

src/fonts/index.scss
a-dg: I had left it out temporarily, as I didn't know if all of the additional character sets (Cyrillic, Greek) were necessary. I'll uncomment this.

src/base/index.scss
a-dg: Good catch, those will be in a separate PR. Thanks!

src/teamsnap-ui.scss
a-dg: Done, I agree that things with no dependencies are better off being imported first.

src/utils/size.scss
a-dg: It seems like anything previously importing `base/index.scss` should import `config/index.scss`, as that's going to pull in the functions and mixins, as well as the new `base/index.scss` to get the reset and typography styles. Does that seem right?

saralohr: Should this be `config/index.scss`?

src/base/reset.scss
```diff
@@ -0,0 +1,22 @@
+@import "../config/index.scss";
```
saralohr: Does this file need to import the config?

src/config/config.scss
```diff
@@ -140,7 +140,7 @@ $su-xxxsmall:   scaleGrid(-3) !default; // 1px
 $su-xxsmall:    scaleGrid(-2) !default; // 2px
 $su-xsmall:     scaleGrid(-1) !default; // 4px
 $su-small:       scaleGrid(1) !default; // 8px
-$su-default:     scaleGrid(2) !default; // 16px
+$su-medium:     scaleGrid(2) !default; // 16px
```
saralohr: 👍 This makes more sense to me. We should probably update in Classic too.

src/base/reset.scss
williamdingwall: not currently, could potentially in the future if htere ever is a specific value needed, unless it's intended purpose will always be to set things to "0".. but just went with it for consistency.

williamdingwall: I'm fine with cutting here

src/config/config.scss
williamdingwall: I can put in a follow up story for that... think there's a few other spots with the functions that need updating in classic too

src/base/reset.scss
saralohr: I guess it's a matter of whether this file (or typography) are being imported directly into the react components? Not sure how these more general base styles are being handled currently, or if they are just relying on Classic for these things. If they are being pulled in in some way, then including it for consistency makes sense...

williamdingwall: I could see a scenario where typography is brought in by itself for various projects... reset less so.  Adding that config in the base files is more about future projects like the marketing site

saralohr: Alrighty, well I'm fine either way. Take it or leave it.

src/utils/size.scss
williamdingwall: ah, yep, updated

PR #23, opened 2017-07-12
Refactor media query mixins to be dynamic

a-dg:


williamdingwall:
@a-dg  I like this update!  A few thoughts...

1) what do you think about simplifying this even further and making the private use function the primary function for media queries, and changing the title to "media-query", so something like
```
@mixin media-query($condition, $width) {
   // content from current @mixin media-query-for-width
  }

usage :
@include media-query(min-width, sm) {
  // added css styles
}
```

2) then changing the current media-query mixin to something like "media-query-aliases", and updating it to work similarly as above so you'd include the min-width or max-width conditions.  Currently I think that function doesn't work with the way we define $breakpoints variable in config, since we don't include that condition anymore.  Would be more flexible too.
```
@mixin media-query-aliases($condition) {
   // content from current @mixin media-query
  }

usage :
@include media-query-aliases(min-width) {
  // added css styles
}
```

3) and updating the comments / table of contents at the top to reflect the new changes

PR #24, opened 2017-07-12
Conform scaled grid values

a-dg:
This forces scaled grid values to be in the following set: 0, 1, 2, 4, 8, and then any multiple of 8. It also adds a new spacing utility variable `$su-xxxsmall`, which represents 1px.

### DIFF
--- a/src/base/config.scss
+++ b/src/base/config.scss
@@ -136,9 +136,9 @@ $tu-large-lineHeight:         scaleFont($lineHeight, 2) !default;
 
 // * 3. Spacing
 
-$su-xxxsmall: scaleGrid(.125) !default; // 1px
-$su-xxsmall:   scaleGrid(.25) !default; // 2px
-$su-xsmall:     scaleGrid(.5) !default; // 4px
+$su-xxxsmall:   scaleGrid(-3) !default; // 1px
+$su-xxsmall:    scaleGrid(-2) !default; // 2px
+$su-xsmall:     scaleGrid(-1) !default; // 4px
 $su-small:       scaleGrid(1) !default; // 8px
 $su-default:     scaleGrid(2) !default; // 16px
 $su-large:       scaleGrid(3) !default; // 24px
--- a/src/base/config.scss
+++ b/src/base/config.scss
@@ -136,6 +136,7 @@ $tu-large-lineHeight:         scaleFont($lineHeight, 2) !default;
 
 // * 3. Spacing
 
+$su-xxxsmall: scaleGrid(.125) !default; // 1px
 $su-xxsmall:   scaleGrid(.25) !default; // 2px
 $su-xsmall:     scaleGrid(.5) !default; // 4px
 $su-small:       scaleGrid(1) !default; // 8px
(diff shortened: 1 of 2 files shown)


williamdingwall:
@a-dg  I like your approach, my only concern is defining those lower values in the function seems to limit some of the flexibility of the function - say if we did decide to update that base unit to 10 or something those lower values would change

what are you're thoughts on something like this?
```
@function scaleGrid($step) {

  // defines base grid spacing unit and rounds step value to enforce
  $spacing: 8px;
  $step: round($step);

  // Handles all positive values
  @if $step >= 1 {
    @return $spacing * $step;
  }

  // converts negative values if $step <= 0
  $step = convertNegative($step)

  // halves base unit based on negative step values
  @for $i from 1 through $step {
    $spacing: $spacing / 2;
  }

  // return spacing if not previous defined with positive step
  @return $spacing;

}
```

a-dg:
That's very clever, much more intuitive than the floats. The example you posted isn't quite working for me for all values, but I get what you're after.

a-dg:
Using your example, is there a reason to accommodate passing `0` to the function? It seems like that should yield `0`, but it somewhat disrupts the pattern, since `1 => 8px` and `-1 => 4px`. What makes sense to you?

williamdingwall:
Yeah I'd agree that making that returned value 0 would disrupt the pattern... having it default to 8 seems a little strange too since 0 and 1 would both equal 8.

maybe let's approach those positive values a little differently so scaleGrid(0) = 8, scaleGrid(1) = 16, etc.  I think that would make sense, since your scaling your grid based on that base value of 8, scaling that base 1 step would be +8; 16px.

maybe something like this?
```
  // defines base grid unit and rounds step value to enforce grid values
  $base-unit: 8px;
  $spacing: $base-unit;
  $step: round($step);

  // Handles all positive values
  @if $step >= 0 {
    @for $i from 1 through $step {
      $spacing: $spacing + $base-unit;
    }
    @return $spacing;
  }
```

a-dg:
I'm going to play devil's advocate and ask if that's being too accommodating for a case that might rarely come up. I'm guessing it's not typical for `scaleGrid()` to be passed a value that resulted from a calculation, correct? So does it make more sense to yield the most intuitive result, even if it doesn't make for an even distribution of values? (Stop me if this is splitting hairs.)

williamdingwall:
Correct, scaleGrid() will likely not be passed something from a calculation, and you're right, there is something intuitive about just multiplying $step & $base-unit to retrieve all positive values. Following that notion I can see the logic in scaleGrid(0) = 0

The way the other functions work is that if passed a 0 value, they just default to the base unit.  So if you're working with these as a suite of functions to cover type, color, and spacing, I do wonder if handling 0 values differently here will make it less intuitive; as well as the negative steps.

... but, I'm cool with just making scaleGrid(0) = 0.  $step * $base-unit would be the most intuitive when trying to think through what step equals a larger value, and that would follow suit.

a-dg:
I'm sorry to belabor this, but I have to backtrack. If `scaleGrid()` needs to be intuitive to use, then passing `0` makes no sense. No developer would do that. I suggest we treat it like the other functions and let it return the base unit, and not make a special case for something that's never going to happen.

PR #25, opened 2017-07-13
Add progress bar component

a-dg:


src/components/ProgressBar.scss
```diff
@@ -0,0 +1,81 @@
...
+    border-radius: ($su-default / 2);
+  }
+}
+
+.ProgressBar-status {
+  height: 100%;
+  position: absolute;
+  top: 0;
+  left: 0;
+  background-color: $cu-primary;
```
williamdingwall: this can be $cu-success for the default

src/components/ProgressBar.scss
```diff
@@ -0,0 +1,81 @@
...
+}
+
+.ProgressBar-range {
+  position: relative;
+  background: $cu-background;
+  height: $su-small;
+  border-radius: ($su-small / 2);
+  margin-top: $su-xsmall;
+  margin-bottom: $su-default;
+  box-shadow: 0 0 2px darken($cu-background, 3%) inset;
```
williamdingwall: This can be updated to use $inset-box-shadow-medium and $inset-box-shadow-small for the smaller progress bars

src/components/ProgressBar.scss
```diff
@@ -0,0 +1,82 @@
...
+// Styles the progress bar for different status
+//---------------------------------------------//
+
+.ProgressBar-status--active,
+.ProgressBar-status--success,
+.ProgressBar-status--waitingForPayment,
+.ProgressBar-status--paid,
+.ProgressBar-status--completed {
+  background: $cu-positive;
+}
```
saralohr: I wonder if these could be simplified to something like `ProgressBar-status--positive`, since these are all applying the same style?

src/components/ProgressBar.scss
```diff
@@ -0,0 +1,82 @@
...
+
+.ProgressBar-status--canceled,
+.ProgressBar-status--failed,
+.ProgressBar-status--danger {
+  background: $cu-negative;
+}
```
saralohr: Same comment here - could this just be `ProgressBar--negative`?

src/components/ProgressBar.scss
```diff
@@ -0,0 +1,82 @@
...
+
+// .ProgressBar-status--completed,
```
saralohr: Same comment for positive and negative styles.

src/components/ProgressBar.scss
saralohr: @williamdingwall did you mean `$cu-positive`? I just ran this locally and I'm getting an error that `$cu-success` is undefined.

a-dg: @saralohr You're right, that was a hasty push on my part.

src/components/ProgressBar.scss
a-dg: My question would be, are there any cases where something other than the color of the progress bar relies on distinguishing between different cases? If not, perhaps what's in Classic is covering a variety of legacy states that aren't needed for this updated version of the component.

saralohr: > Are there any cases where something other than the color of the progress bar relies on distinguishing between different cases?

Not that I'm aware of.

> Perhaps what's in Classic is covering a variety of legacy states that aren't needed for this updated version of the component.

Yes, I think this is the case.

williamdingwall: yep, that is the case.  I like the idea of simplifying to positive, negative, and neutral.

williamdingwall:
@a-dg  any luck with the Radial progress bar?

src/components/ProgressBar.scss
williamdingwall: @sara Yep, my bad.

a-dg:
@williamdingwall I haven't yet worked on the radial version, will start on that next week.

captain-pj:
New styles for progress bar component.

src/components/ProgressBar.scss
```diff
@@ -0,0 +1,70 @@
+// @import "../base/index.scss";
```
williamdingwall: You can delete this whole file since there's the new progress.scss file now

src/components/progress.scss
```diff
@@ -0,0 +1,133 @@
...
+  border-radius: $border-radius-large;
+  margin-bottom: $su-small;
+  .ProgressBar--large & {
+    height: scaleGrid(2);
+  }
+  .ProgressBar--small & {
+    height: scaleGrid(.5);
+  }
+  .ProgressBar--extra-small & {
+    height: scaleGrid(.25);
```
williamdingwall: This should be `scaleGrid(-2)` now

src/components/progress.scss
```diff
@@ -0,0 +1,133 @@
...
+  flex: 1; // for inline style
+  background: $cu-middleground--dark;
+  box-shadow: inset $box-shadow-small;
+  border-radius: $border-radius-large;
+  margin-bottom: $su-small;
+  .ProgressBar--large & {
+    height: scaleGrid(2);
+  }
+  .ProgressBar--small & {
+    height: scaleGrid(.5);
```
williamdingwall: This should be `scaleGrid(-1)`

src/components/progress.scss
```diff
@@ -0,0 +1,133 @@
...
+        clip: rect(auto, auto, auto, auto)
+      }
+    }
+
+  }
+  &.ProgressBar--small .ProgressBar-range {
+    width: scaleGrid(4);
+    height: scaleGrid(4);
+    &:before {
+      border-width: scaleGrid(.5);
```
williamdingwall: This should be `scaleGrid(-1)

src/components/progress.scss
```diff
@@ -0,0 +1,133 @@
...
+    }
+    .ProgressBar-status {
+      border-width: scaleGrid(.5);
```
williamdingwall: This should be `scaleGrid(-1)`

src/components/progress.scss
```diff
@@ -0,0 +1,133 @@
...
+  }
+}
+
+// radial progress bar
+.ProgressBar--radial {
+  position: relative;
+  display: block;
+  height: auto;
+  width: 100%;
+  margin-bottom: $su-meduim;
```
williamdingwall: this should be `$su-default`

src/components/ProgressBar.scss
captain-pj: I had deleted that in a previous commit, but I think with the rebasing, we deleted the wrong one yesterday. The same applies to the notes below where I actually changed most of the scaleGrid(x) to $su-size variables so that it would be more consistent.

captain-pj:
Pushed out lost changes from yesterday. Let me know if all of this looks better!

src/components/progress.scss
```diff
@@ -0,0 +1,133 @@
...
+  margin-bottom: $su-xsmall;
+}
+
+.ProgressBar-status {
+  background: $cu-positive;
+  height: 100%;
+  border-radius: $border-radius-large;
+  &.ProgressBar-status--canceled,
+  &.ProgressBar-status--failed,
+  &.ProgressBar-status--danger {
```
saralohr: Could we reduce this to one class? Like `ProgressBar-negative`

src/components/progress.scss
```diff
@@ -0,0 +1,133 @@
...
+    background: $cu-negative;
+  }
+  &.ProgressBar-status--authorized,
+  &.ProgressBar-status--chargedBack,
+  &.ProgressBar-status--refunded {
```
saralohr: Same question here, can these be reduced to one class?

src/components/progress.scss
```diff
@@ -0,0 +1,133 @@
+// * Progress Bar component
```
saralohr: I wonder if this should be capitalized `Progress` or `ProgressBar` to be consistent with the other components. There are a few that aren't capped but these are legacy from prior to when we changed the convention for to prevent conflicts when importing into LFE.

src/components/progress.scss
```diff
@@ -0,0 +1,133 @@
...
+  box-shadow: inset $box-shadow-small;
+  border-radius: $border-radius-large;
+  margin-bottom: $su-small;
+  .ProgressBar--large & {
+    height: $su-default;
+  }
+  .ProgressBar--small & {
+    height: $su-xsmall;
+  }
+  .ProgressBar--extra-small & {
```
saralohr: These modifiers should be camel case or perhaps reduced to one word. https://teamsnap-ui-patterns.netlify.com/docs/naming-convention.html

ie `ProgressBar--extraSmall` or if it's treated as just one word it could be `ProgressBar--xsmall`

src/components/progress.scss
captain-pj: These classes were already in the component, so I just used them. If it's okay to drop the others, I am all for it!

src/components/progress.scss
captain-pj: Not sure exactly what you mean here. Just change it in the comment? I tried to follow the pattern of the others, but let me know what I should change it to.

saralohr: Sorry.. I just meant the file name.

src/components/progress.scss
captain-pj: Oh gotcha! I think I copied over the naming style from an old component. I will keep this in mind going forward. I will do xsmall and xlarge!

src/components/progress.scss
saralohr: Yeah, see comment from when Alan was working on this: https://github.com/teamsnap/teamsnap-ui/pull/25#pullrequestreview-49813369

saralohr: Whoops that link didn't work as expected... try this: https://github.com/teamsnap/teamsnap-ui/pull/25#discussion_r127248001

captain-pj: Sounds good. In that case, do we need "ProgressBar--positive?" It seems from the style guide that the positive color is the default.

src/components/progress.scss
captain-pj: Got it!

src/components/progress.scss
saralohr: Sorry -- missed this comment. @williamdingwall might be better able to answer this question.

williamdingwall: No we don't think we do since it's the default.  By default all progress within are system is positive. Unless specified with one of the two modifiers, `ProgressBar--negative` and `ProgressBar--neutral`

PR #27, opened 2017-08-03
Add button group component

saralohr:
button-group

### DIFF
--- a/src/components/ButtonGroup.scss
+++ b/src/components/ButtonGroup.scss
@@ -1,0 +1,35 @@
+// * ButtonGroup
+
+$ButtonGroup-border-radius: $border-radius-medium;
+$ButtonGroup-space: -1 * $border-width-small;
+
+
+// * 1. Lay out items left to right
+// * 2. Make IE10 play nice
+
+.ButtonGroup {
+  display: flex; // 1
+  display: -ms-flexbox; // 2
+}
+
+
+// * 1. Overlap elements and override any vertical margins.
+// * 2. Reset `border-radius` so corners line up.
+
+.ButtonGroup > * {
+  margin: 0 $ButtonGroup-space; // 1
+  border-radius: 0; // 2
+}
+
+
+// * Apply border-radius outer edges of first and last children
+
+.ButtonGroup > *:first-child {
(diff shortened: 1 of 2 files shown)


src/components/Button.scss
```diff
@@ -50,8 +50,6 @@ $Button-padding: 0 scaleGrid(2);
 
   &.is-active {
     background-color: $Button-bg-hover;
-    border-bottom-width: 1px;
-    border-top-width: $Button-border-bottom-width;
   }
```
saralohr: @williamdingwall What do you think about this change? Adding the heavy border to the top on active state always seemed a little weird to me... especially in the button group, but maybe that's just me!

src/components/ButtonGroup.scss
```diff
@@ -0,0 +1,35 @@
...
+.ButtonGroup {
+  display: flex; // 1
+  display: -ms-flexbox; // 2
+}
+
+
+// * 1. Overlap elements and override any vertical margins.
+// * 2. Reset `border-radius` so corners line up.
+
+.ButtonGroup > * {
```
williamdingwall: Couldn't this just be .ButtonGroup > .Button, I'm guessing the button class will always be the direct children of button right?

src/components/ButtonGroup.scss
```diff
@@ -0,0 +1,35 @@
...
+  margin: 0 $ButtonGroup-space; // 1
+  border-radius: 0; // 2
+}
+
+
+// * Apply border-radius outer edges of first and last children
+
+.ButtonGroup > *:first-child {
```
williamdingwall: Same here

src/components/ButtonGroup.scss
```diff
@@ -0,0 +1,35 @@
...
+  border-top-left-radius: $ButtonGroup-border-radius;
+  border-bottom-left-radius: $ButtonGroup-border-radius;
+}
+
+.ButtonGroup > *:last-child {
```
williamdingwall: and here

src/components/ButtonGroup.scss
saralohr: Yep, prolly so! I'll change this.

PR #29, opened 2017-08-08
Adding loading and divider components

williamdingwall:


src/components/Divider.scss
```diff
@@ -0,0 +1,38 @@
+//------------------------------------------------------------//
+
+// * Dividers
+// * 1. chapter
+// * 2. Idented Divider
```
saralohr: Spelling nit: "Indented"

src/components/Divider.scss
```diff
@@ -0,0 +1,38 @@
+//------------------------------------------------------------//
+
+// * Dividers
+// * 1. chapter
```
saralohr: What does chapter refer to?

williamdingwall: Oh, nothing, I just have a snippet saved as a template for my CSS comments and was rushing at the end of this.
```

// * Title
// * 1. chapter

// Notes:
// * none

// TODO : none

// * 1. Chapter

```

src/components/Loader.scss
```diff
@@ -0,0 +1,248 @@
...
+  100% {
+    #{$prefix}transform: translateY(0) rotate(90deg);
+  }
+}
+
+@keyframes jelloBox {
+  @include jelloBox('');
+}
+
+@-webkit-keyframes jelloBox {
```
saralohr: I _think_ we are using Autoprefixer? We shouldn't need to add prefixes within the CSS.

saralohr: Just checked: we're [prefixing](https://github.com/teamsnap/ui-patterns/blob/master/gulpfile.js#L43) in `ui-patterns` and I believe @Ollo mentioned LFE had Autoprefixer in their pipeline too.

williamdingwall: Oh, yep, we are : https://github.com/teamsnap/league-frontend/blob/2890c7aef77d61af0064c2941da9f99650b51f26/build/webpack.config.js#L230

I'll cut

src/components/Loader.scss
```diff
@@ -0,0 +1,248 @@
...
+//------------------------------------------------------------//
+
+// * 1. Loading Indicators
+
+// * 1.a Jello Indicator
+
+.JelloAnimation {
+  display: inline-block;
+  position: relative;
+  height: scaleGrid(5);
```
saralohr: If you gave this a height too, maybe this could act more like a block element?

src/components/Loader.scss
```diff
@@ -0,0 +1,248 @@
...
+
+@-webkit-keyframes spin {
+  @include SpinAnimation('-webkit-');
+}
+
+// * 2 Page Loader
+
+.PageLoader {
+  width: 100%;
+  height: 125px;
```
saralohr: Why the explicit `height`?

src/components/Loader.scss
```diff
@@ -0,0 +1,248 @@
...
+  margin-bottom: $su-small;
+}
+
+.PageLoader-indicatorAnimation {
+  position: absolute;
+  top: 0;
+  bottom: 0;
+  left: 0;
+  margin: auto;
+}
```
saralohr: I wonder if `PageLoader-indicator` and `PageLoader-indicatorAnimation` could be simplified? Seems like a lot going on to center two items side by side. Could `display: flex` with `justify-content: center` work, replacing all the absolute positioning?

src/components/Loader.scss
```diff
@@ -0,0 +1,248 @@
...
+  color: $cu-info--light;
+}
+
+// * 2.a Page Loader Modifiers
+
+.PageLoader--centered {
+  position: fixed;
+  top: 50%;
+  left: 0;
+  margin-top: -63px;
```
saralohr: This might be able to get there without pixel values using `transform: translateY(-50%)`.

src/components/Loader.scss
```diff
@@ -0,0 +1,211 @@
...
+
+}
+
+// * 3 Block Loader
+
+.BlockLoader {
+  display: block;
+  text-align: center;
+  padding: $su-small 0;
+}
```
saralohr: This feels really similar to the basic `PageLoader`.. what's the difference?

src/components/Loader.scss
williamdingwall: Yeah, good call

src/components/Loader.scss
williamdingwall: Yeah I guess we could probably simplify both loaders down to a single loader, that just has elements that can be added on

williamdingwall:
Ha, thanks!  and Agreed,  went in an cleaned up that loader component a bit to be a single component... Good calls all the around with the review comments!

PR #31, opened 2017-08-30
Adds missing import in Loader.scss

Ollo:
Adds the missing `@import` and bumps package to 0.5.1


mikereinmiller:
Thanks for jumping on this @Ollo Looks good 👍

PR #32, opened 2017-09-06
Responsive Panel and utilities

saralohr:
+ Extends `.Panel` component to account for table-like layouts and responsive layouts
+ Add base and responsive utilities for text, display, space and borders

### DIFF
--- a/src/components/Panel.scss
+++ b/src/components/Panel.scss
@@ -5,6 +5,7 @@
 // * Panels
 // * 1. Panel Settings
 // * 2. Base Panel
+// * 3. Panel Modifiers
 
 // Notes:
 // * also add to TeamSnap-UI sketch doc and create some rules
@@ -12,7 +13,6 @@
 
 // TODO : Create embedable version of panel for demos
 // TODO : Add modifier for panel with button in header
-// TODO : look into panel with table body; do we need it?
 // TODO : revisit theming and how that relates to panel styles
 // TODO : revisit panel-footer alignment pattern
 // TODO : Refactor containers and header in classic; conflicts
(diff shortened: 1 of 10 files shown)


src/utils/border.scss
```diff
@@ -0,0 +1,36 @@
...
+  }
+  .u-border#{$name}None {
+    border-#{$value}: 0;
+  }
+  .u-borderNone {
+    border: 0;
+  }
+}
+
+@if $responsive == true {
```
saralohr: Since the border logic is now being handled within the `Panel--stacked` class, the responsive utilities aren't really necessary yet, but if they seem useful we can keep them in.

saralohr:
Note to self and @williamdingwall: This should be checked against existing `Panel` usage in LFE.

mikereinmiller:
Good to know, thanks @saralohr! We'll have to make a note to test for this when we upgrade the UI version in LFE.

src/components/Panel.scss
```diff
@@ -35,36 +35,121 @@ $Panel-header-text-color: $cu-info !default;
...
-    border-bottom: 0;
-  }
+// Override default <table> cell styles for when table markup is used
+.Panel-cell {
+  border-top: 0;
+  border-bottom: 0;
+}
 
+.Panel-cell--header {
+  background-color: $cu-background--light;
```
williamdingwall: Same here... just anywhere anywhere it's currently used here... Since background--light === background, $cu-middleground is probably the one you were looking for.

src/teamsnap-ui.scss
```diff
@@ -1,3 +1,5 @@
+$responsive: true;
```
williamdingwall: Hmmm... let's just set it as the default in the config instead. Rather than setting it as false in the config and overriding for everywhere.

_Follow up :_ we might need to think more about how we turn the responsiveness off and on when we start implementing in the app too.  Just leaving it on is fine for now and people can just use the appropriate classes, but when we get to a point where we want to start using those responsive classes and turn the responsiveness of those classes off and on, just generating or not generating the class won't work.  I think we'll have to add the classes regardless and set the breakpoints values to 0 like we do in classic : https://github.com/teamsnap/classic/blob/master/app/assets/stylesheets/framework/settings/_layout.scss#L34

src/utils/border.scss
williamdingwall: I think there's use in keeping them

src/teamsnap-ui.scss
saralohr: @williamdingwall  I'm not totally following this statement:

> when we get to a point where we want to start using those responsive classes and turn the responsiveness of those classes off and on

Let's discuss soon. It's okay to leave as-is for the moment, though?

src/components/Panel.scss
```diff
@@ -35,36 +35,121 @@ $Panel-header-text-color: $cu-info !default;
...
 .Panel-footer {
   border-top: 1px solid $cu-divider;
   text-align: right;
 }
+
+
+// * 3. Modifiers
+
+.Panel--striped .Panel-row:nth-child(even) {
+  background-color: $cu-background--light;
```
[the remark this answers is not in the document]
saralohr: Okay, I'll switch to `middleground` for now and we can re-evaluate these variables. I'll add a ticket for this.

src/teamsnap-ui.scss
williamdingwall: Yeah that's fine, we can talk monday.. and I'll update in the re-organize story

PR #33, opened 2017-09-08
Refactor media query mixins to be dynamic

williamdingwall:
condensing work from these two PRs..

https://github.com/teamsnap/teamsnap-ui/pull/23
https://github.com/teamsnap/teamsnap-ui/pull/24

..and wrapping up

### DIFF
--- a/src/base/config.scss
+++ b/src/base/config.scss
@@ -136,8 +136,9 @@ $tu-large-lineHeight:         scaleFont($lineHeight, 2) !default;
 
 // * 3. Spacing
 
-$su-xxsmall:   scaleGrid(.25) !default; // 2px
-$su-xsmall:     scaleGrid(.5) !default; // 4px
+$su-xxxsmall:   scaleGrid(-3) !default; // 1px
+$su-xxsmall:    scaleGrid(-2) !default; // 2px
+$su-xsmall:     scaleGrid(-1) !default; // 4px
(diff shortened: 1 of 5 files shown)


src/base/config.scss
```diff
@@ -136,8 +136,9 @@ $tu-large-lineHeight:         scaleFont($lineHeight, 2) !default;
 
 // * 3. Spacing
 
-$su-xxsmall:   scaleGrid(.25) !default; // 2px
-$su-xsmall:     scaleGrid(.5) !default; // 4px
+$su-xxxsmall:   scaleGrid(-3) !default; // 1px
+$su-xxsmall:    scaleGrid(-2) !default; // 2px
+$su-xsmall:     scaleGrid(-1) !default; // 4px
```
saralohr: I like this change... I wasn't following the whole previous `scaleGrid` thread, but this treatment feels more intuitive to me.

src/base/mixins.scss
```diff
@@ -1,159 +1,23 @@
...
+// * none
 
 // TODO : look into newer SASS media queries further
-// TODO : look into simplifying media-query
+// TODO : create mixin wrapper for responsive aliases
 
 //------------------------------------------------------------//
 
-// * 1. media-query()
-
```
saralohr: This makes me happy.

PR #34, opened 2017-09-13
DES-96 Initial styling of list summary

captain-pj:


src/components/SummaryList.scss
```diff
@@ -0,0 +1,42 @@
+.SummaryList-header {
+  color: $base-font-color;
```
saralohr: I wonder if this element could be simplified a bit. Maybe you could use flex on the parent and then the children wouldn't need classes or floats. Something like:
```
.SummaryList-heading {
  color: $base-font-color;
  font-size: $tu-large-fontSize;
  display: flex;  
  justify-content: space-between;
  margin-bottom: $su-xsmall;
}

<h3 class="SummaryList-heading">
  <span>Invoice</span>
  <span>Amount</span>
</h3>
```

PS:
+ We have a greedy attribute selector already for `[class*="-header"]` that applies a pseudo element to clear floats. This messes up the flex display because it considers this an equal child and tries to give it space. One work around is to add `flex-grow: 1` to a child. Another workaround is to just not use `header`... hence `heading` above. :eye-roll:
+  I added a font-size so that any heading element can be used and will display the same, unless we _want_ the size to be changeable.

src/components/SummaryList.scss
```diff
@@ -0,0 +1,42 @@
...
+    float: left;
+  }
+  .SummaryList-header-value {
+    color: $base-font-color;
+    float: right;
+    clear: right;
+  }
+}
+
+.SummaryList-row {
```
saralohr: Nit: The name row here feels a little off to me... the list children feel more like the rows... maybe it could be something more generic like `SummaryList-section`?

src/components/SummaryList.scss
```diff
@@ -0,0 +1,42 @@
...
+  border-top: $border-width-large solid $ts-grey;
+  padding: $su-default 0;
+  list-style-type: none;
+  li {
```
saralohr: I'd like to see this element have a class name and be un-nested from the parent to reduce specificity . `SummaryList-item` or similar.

src/components/SummaryList.scss
```diff
@@ -0,0 +1,42 @@
...
+    display: flex;
+    width: 100%;
+    height: 30px;
+    align-items: center;
+    .SummaryList-item-value {
```
saralohr: Nit: Should be camel case `SummaryList-itemValue` and doesn't need to be nested under parent.

src/components/SummaryList.scss
```diff
@@ -0,0 +1,42 @@
...
+      display: flex;
+      flex-grow: 1;
+      &:before {
+        content: ' ';
+        flex-grow: 1;
+        margin-bottom: 4px;
+        border-bottom: $border-width-small dashed $ts-grey;
+        width: 10%;
+      }
```
saralohr: Love this treatment for the dot leader!!

src/components/SummaryList.scss
saralohr: Would a modifier like this work for our bordered instances?
```
SummaryList-item--bordered:not(:last-child) {
  border-bottom: [style];
}
```

src/components/SummaryList.scss
captain-pj: I took the classes off of these and just put the styles on the spans within the SummaryList-item above

src/components/SummaryList.scss
captain-pj: :)

src/components/SummaryList.scss
captain-pj: Used your example above and changed the class name.

src/components/SummaryList.scss
captain-pj: Yeah, good call. Changed to section.

PR #35, opened 2017-09-14
DES-99 initial styling for errors

captain-pj:
I have a few questions in the scss comments about this. I did the styling and created some elements in the components for errors/feedback but am wondering if I should create a form component so it has examples of all possiblities (option/checkbox/radio/etc.) Also had some questions I posed to slack about the use of svg icons within the css.

### DIFF
--- a/src/components/Input.scss
+++ b/src/components/Input.scss
@@ -7,6 +7,7 @@
 // * 2. Base Input
 // * 3. Inline Modifiers
 // * 4. Amount and Calendar Modifier
+// * 5. Errors and feedback
 
 // Notes:
 // * size utilities should be used on the Input-element to
@@ -66,7 +67,7 @@ $form-input-bg-color: $cu-foreground--dark !default;
   padding: 0;
 }
 
-.is-required {
+.Input.is-required {
   .Input-label:after {
     content: '*';
     display: inline-block;
@@ -89,7 +90,7 @@ $form-input-bg-color: $cu-foreground--dark !default;
   text-align: right;
 }
 
-.is-required {
+.Input.is-required {
   .Input-label--inline:after {
     display: none;
   }
@@ -158,5 +159,39 @@ $form-input-bg-color: $cu-foreground--dark !default;
       @include ss-pika;
     }
   }
+}
 
+// * 5. Errors and feedback
+.Input-feedback {
+  display: none;
+}
+.Input.is-required.is-invalid {
+  .Input-label:after {
+    color: $cu-negative;
+  }
+}
+.Input.is-required.is-valid {
+  .Input-label:after {
+    content: ' ';
+  }
+}
+.Input.is-invalid {
+  .Input-element {
+    border-color: $cu-negative;
+  }
+  .Input-feedback {
+    display: block;
+    color: $cu-negative;
+  }
+}
+.Input.is-valid {
+  .validation {
+    fill: $cu-positive;
+    position: absolute;
+    top: $su-xxsmall;
+    left: 0;
+  }
+  .Input-label {
+    padding-left: $su-medium;
+  }
 }


src/components/Input.scss
```diff
@@ -158,5 +159,39 @@ $form-input-bg-color: $cu-foreground--dark !default;
       @include ss-pika;
     }
   }
+}
 
+// * 5. Errors and feedback
+.Input-feedback {
+  display: none;
+}
+.is-invalid {
```
saralohr: States _should_ always be attached to a specific class, since they will be repeated frequently:
https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md#componentnameis-stateofcomponent

src/components/Input.scss
```diff
@@ -158,5 +159,39 @@ $form-input-bg-color: $cu-foreground--dark !default;
...
+    border-color: $cu-negative;
+  }
+  .Input-feedback {
+    display: block;
+    color: $cu-negative;
+  }
 }
+// should this be moved to its own scss file since it will also apply to checkboxes/radios/selects?
+// should there be a "forms" component with examples like this and its own scss?
+.is-valid {
```
saralohr: We should probably style the different form components individually with error states. If we want to provide a more generic treatment this could be added via some sort of `FormControl` wrapper.

src/components/Input.scss
```diff
@@ -158,5 +159,39 @@ $form-input-bg-color: $cu-foreground--dark !default;
...
+    .Input-label:after {
+      color: $cu-positive;
+      //content: url(assets/icons/check.svg); <- is this folder included in any build path?
+    }
+  }
+  .Input-label:after {
+    color: $cu-positive;
+    display: inline-block;
+    padding-left: $su-xsmall;
+    content: '#'; // just a hash until I get clarity on the svg paths
```
saralohr: I'd use SVG if you can. Not sure how the path generation will work though. We don't have anything built for that and it may get complicated across different app implementations.

src/components/Input.scss
captain-pj: Updated that. ".is-invalid" is kind of on its own earlier in the file, so I was trying to follow the pattern there. Let me know if this works any better.

src/components/Input.scss
captain-pj: Used svg in the component rather than using it here in the css. Looking around it seems like that is the better way as SVG referenced in a uri in css can really slow loading. Let me know in the patterns library if you think that is a good way to go for that.

src/components/Input.scss
```diff
@@ -158,5 +159,39 @@ $form-input-bg-color: $cu-foreground--dark !default;
       @include ss-pika;
     }
   }
+}
 
+// * 5. Errors and feedback
+.Input-feedback {
+  display: none;
+}
+.is-required.is-invalid {
```
saralohr: I'd still like to see these classes attached to whatever class they are modifying.

src/components/Input.scss
```diff
@@ -158,5 +159,39 @@ $form-input-bg-color: $cu-foreground--dark !default;
...
+  .Input-label:after {
+    color: $cu-negative;
+  }
 }
+.is-required.is-valid {
```
saralohr: Same here.

src/components/Input.scss
captain-pj: Should I update elsewhere in the file as well? Like here: 

.is-required {
  .Input-label--inline:after {
    display: none;
  }
  .Input-label--inline:before {
    content: '*';
    display: inline-block;
    padding-right: $su-xsmall;
    color: $cu-info--light;
  }
}

And here:

.is-required {
  .Input-label:after {
    content: '*';
    display: inline-block;
    padding-left: $su-xsmall;
    color: $cu-info--light;
  }
}

src/components/Input.scss
```diff
@@ -158,5 +159,39 @@ $form-input-bg-color: $cu-foreground--dark !default;
...
+  .Input-element {
+    border-color: $cu-negative;
+  }
+  .Input-feedback {
+    display: block;
+    color: $cu-negative;
+  }
+}
+.Input.is-valid {
+  svg {
```
saralohr: Can we add a class to this instead of referencing the element directly? I could see a scenario where there might be other SVG within an Input component.

src/components/Input.scss
saralohr: Yeah, these should all be modifying a class. If you want to update it that would be great.

src/components/Input.scss
```diff
@@ -158,5 +159,39 @@ $form-input-bg-color: $cu-foreground--dark !default;
...
+}
+.is-required.is-invalid {
+  .Input-label:after {
+    color: $cu-negative;
+  }
 }
+.is-required.is-valid {
+  .Input-label:after {
+    content: ' ';
+  }
```
saralohr: What is this doing?

captain-pj: It's removing the is-required star from the form if the element is marked with is-valid.

https://teamsnap.invisionapp.com/share/QTBYOK53R#/screens/247877030

The stars aren't there when the form element is valid. Also, the is-required star in the style guide is before the label element. As it exists now, it is built on the :after of the label. Let me know if that is not okay and I can correct that as well.

saralohr: Ah, I see. Yeah, I guess the `:before` `:after` is a tad confusing, but often these aren't used to position things in any order so it doesn't bother me too much. Totally fine if you want to change that, though. Could you just add a line of documentation here to say what the `content: ''` is doing? Since it is out of context from the star placement, it is not obvious.

PR #36, opened 2017-09-26
initial feedback styling and update to var in summary list causing co…

captain-pj:
…nflicts

Want some initial feedback on what I have been working on with the feedback and notifications.

### DIFF
--- a/src/css/components/Feedback.scss
+++ b/src/css/components/Feedback.scss
@@ -1,0 +1,120 @@
+@import "../config/index.scss";
+
+//------------------------------------------------------------//
+
+// * Feedback Component
+// * 1. Define Default FeedBack
+// * 2. Color Modifiers
+// * 3. Type Modifiers
+
+// * Notes:
+// * 1: uses fontSize variable to lineup x-height with width
+
+//------------------------------------------------------------//
+
+// * 1. Define Default FeedBack
+
+.Feedback {
+  position: relative;
+  width: 100%;
+  border: 1px solid $cu-primary;
+  border-radius: $border-radius-medium;
+  margin: $su-small 0;
+  padding: $su-small;
+  background: $ts-white;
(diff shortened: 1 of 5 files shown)


captain-pj:
Updated the general design. Since I was going off the original screengrab in the ticket, I wasn't sure if we needed all of the color possibilities or not. Let me know if it is still too colorful or any other feedback you might have.

williamdingwall:
**Re-wrote a lot of this and updated the progress component**

src/css/components/Feedback.scss
```diff
@@ -0,0 +1,120 @@
+@import "../config/index.scss";
```
saralohr: Do we need to do this still? I think we were doing this for the sake of CSS modules in LFE.

src/css/components/ProgressBar.scss
```diff
@@ -1,135 +1,301 @@
...
+    &, .ProgressBar-status {
+      border-radius: scaleGrid(8) / 2;
     }
+  }
+
+}
+
+// * 3. Precise Modifier
+
+.ProgressBar.ProgressBar--precise {
```
saralohr: Can this just be `.ProgressBar--precise`?

williamdingwall: needed it for specificity

src/css/components/Feedback.scss
williamdingwall: yeah we could probably drop it - just keeping this file consistent with the rest

src/css/components/ProgressBar.scss
williamdingwall: ..actually nevermind, updated it. Was running into a specificty issue but not an issue. also moved this to a new PR : https://github.com/teamsnap/teamsnap-ui/pull/74

PR #39, opened 2017-09-28
Update Panel CSS

saralohr:
+ Include styles for header button so header height doesn't increase when button is added
+ Include flex properties on rows with cells so these don't always have to be added via markup
+ Write better documentaion

### DIFF
--- a/src/components/Panel.scss
+++ b/src/components/Panel.scss
@@ -219,6 +219,7 @@ $Panel-header-text-color: $cu-info !default;
 
         .Panel-row {
           flex-wrap: wrap; // 1
+          border-bottom: 0;
         }
 
         .Panel-cell {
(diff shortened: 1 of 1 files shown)


src/components/Panel.scss
```diff
@@ -228,7 +229,12 @@ $Panel-header-text-color: $cu-info !default;
...
         .Panel-cell--header {
           border-right: 0; // 3
-          border-bottom: 1px solid $cu-divider--light;  // 3
+          padding-bottom: 0;
+          background-color: inherit;
+        }
+
+        .Panel-cell--header + .Panel-cell {
+          padding-top: $su-small;
         }
```
saralohr: We simplified the design of this modifier. It was designed for use in two-column tables, and is currently being battle tested in the account section. The cell headers background color and borders were feeling too busy, so we made the cell headers act more like simple text labels.

stacked-account


stacked

PR #40, opened 2017-09-28
Update border, flex and display utilities:

saralohr:
+ Update border scss loop so there is not repetition of u-border and u-borderNone
+ Move u-flex to display so that it can be used in succession with other responsive display utilities

### DIFF
--- a/src/utils/border.scss
+++ b/src/utils/border.scss
@@ -7,31 +7,49 @@ $borders: (
   'Top': top
 );
 
+$border-style: 1px solid $cu-divider--light;
+
+.u-border {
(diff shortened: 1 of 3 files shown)


src/utils/border.scss
```diff
@@ -7,31 +7,49 @@ $borders: (
   'Top': top
 );
 
+$border-style: 1px solid $cu-divider--light;
```
williamdingwall: since we have a standard and slightly lighter border, think we need differentiate between the two here?  or will this be specific to use within tab panels/tables?

PR #43, opened 2017-10-10
first stab at calendar component versions

captain-pj:


src/components/Calendar.scss
```diff
@@ -0,0 +1,162 @@
...
+  background: $cu-background--light;
+  margin: 0;
+  padding: $su-small 0;
+  border: 1px solid $ts-grey;
+  color: $cu-primary;
+  list-style-type: none;
+}
+.Calendar-row[data-ui-component="week"] {
+  border-right: 1px solid $ts-grey;
+  & > [class*="Calendar-cell"] {
```
saralohr: Does this guy need to be nested? 

What the thinking behind the everything selector?

src/components/Calendar.scss
```diff
@@ -0,0 +1,162 @@
...
+    height: auto;
+    padding-bottom: $su-small;
+    border-left: 1px solid $ts-grey;
+    border-bottom: 1px solid $ts-grey;
+    span {
+      display: block;
+    }
+  }
+  & > [data-ui-component="current-day"] {
```
saralohr: Do these guys need to be nested? Could they just stand alone as cell modifiers, ie `Calendar-cell--today`, `Calendar-cell--nextMonth`?

src/components/Calendar.scss
```diff
@@ -0,0 +1,162 @@
...
+    background: $cu-background--light;
+    span, abbr {
+      opacity: 0.6;
+    }
+  }
+}
+
+// responsive/widget
+
+.Calendar-widget--exampleWrapper {
```
saralohr: I don't think this needs to be part of the component -- the width will likely be derived from a layout wrapper. 

To get this width for the purpose of the demo, you could just add inline styles.

src/components/Calendar.scss
```diff
@@ -0,0 +1,162 @@
...
+  width: 300px;
+}
+
+.Calendar.Calendar-widget {
```
saralohr: This should just be a standalone modifier: `Calendar--widget`.

src/components/Calendar.scss
```diff
@@ -0,0 +1,162 @@
...
+    span, abbr {
+      display: none;
+    }
+  }
+  }
+  .Calendar-row {
+    box-sizing: border-box;
+    display: flex;
+    -webkit-box-orient: horizontal;
+    -webkit-box-direction: normal;
```
saralohr: We don't need to provide browser prefixes since autoprefixer is part of the build process.

src/components/Calendar.scss
```diff
@@ -0,0 +1,162 @@
...
+    flex: 1 1 0%;
+    box-sizing: inherit;
+    min-width: 0;
+    padding: $su-medium;
+  }
+}
+
+// facebook/pagination button - maybe break this out to the buttons section? Also- need an f icon?
+
+.Button--facebook {
```
saralohr: Yeah, this should go with `Button`s.

src/components/Calendar.scss
```diff
@@ -0,0 +1,162 @@
...
+    &:hover,
+    &:active {
+      background-color: $cu-primary;
+      border-color: $cu-primary--dark;
+      color: $ts-white;
+    }
+  }
+}
+
+.Button--paginate {
```
saralohr: I don't know if we need a Button modifier just for increasing font weight. It would come in handy to have a _text_ modifier to increase font-weight instead. That way this property could be used anywhere we need it. 

This could be added in the text utilities. Something like `u-textBold`.

src/components/Calendar.scss
```diff
@@ -0,0 +1,162 @@
...
+    text-align: center;
+  }
+  .Calendar-row[data-ui-component="week"] > [class*="Calendar-cell"] {
+    height: 150px;
+    padding-bottom: 0;
+  }
+  .Calendar-row[data-ui-component="week-heading"] li {
+    display: block;
+  }
+  .Calendar-row {
```
saralohr: Since these styles are repeated both here and for `Calendar--widget`, I wonder if these don't just become our default styles for `Calendar-row` and `Calendar-cell`. For the responsive version we could add those styles in a `max-width` media query instead. It's not an ideal mobile first solution, but I think it might make more sense given our current implementations.

src/components/Calendar.scss
saralohr: Please add all the base `Calendar` styles first and then add parent-component-level modifiers (ie `Calendar--widget`) at the bottom. It's a bit easier to follow that way and is consistent with how other components are written. Modifiers of children are fine to include in the main flow.

src/components/Calendar.scss
```diff
@@ -0,0 +1,162 @@
...
+        float: right;
+      }
+    }
+
+  }
+  .Calendar-row[data-ui-component="week-heading"] li {
+    display: block;
+  }
+  .Calendar-row[data-ui-component="week"] {
+    & > [class*="Calendar-cell"] {
```
saralohr: Remove nesting if possible.

src/components/Calendar.scss
```diff
@@ -0,0 +1,162 @@
...
+  }
+  .Calendar-row[data-ui-component="week-heading"] li {
+    display: block;
+  }
+  .Calendar-row {
+    box-sizing: border-box;
+    display: flex;
+    -webkit-box-orient: horizontal;
+    -webkit-box-direction: normal;
+    flex-flow: row wrap;
```
saralohr: Do we need the `wrap` property here? Doesn't seem like we actually want the cells to wrap. Since the default flex direction is row, I think we can leave out this whole property.

src/components/Calendar.scss
```diff
@@ -0,0 +1,162 @@
+.Calendar-heading, .Calendar-secondary-heading {
+  height: 60px;
```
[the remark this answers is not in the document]
saralohr: Another thought.. not sure the `Calendar-heading` really needs to be a part of this component. This seems more specific to the layout surroundings than the calendar itself. The `Calendar-secondaryHeading` feels more like the main heading to me, specific and integral to the calendar.

@williamdingwall may have thoughts on this in regards to current usage. Obviously you don't have to include the `Calendar-heading`, but then the `secondary` feels a little off to me.

src/components/Calendar.scss
saralohr: `Calendar-row` needs to have a `max-width: none`  applied to it to subvert the universal styles applied to `[class*=-row]` that set a max-width of 1300px.

src/components/Calendar.scss
```diff
@@ -0,0 +1,162 @@
...
+  list-style-type: none;
+}
+.Calendar-row[data-ui-component="week"] {
+  border-right: 1px solid $ts-grey;
+  & > [class*="Calendar-cell"] {
+    height: auto;
+    padding-bottom: $su-small;
+    border-left: 1px solid $ts-grey;
+    border-bottom: 1px solid $ts-grey;
+    span {
```
saralohr: Can this be written as a block level element instead, like a `<p>`?

src/components/Calendar.scss
saralohr: screen shot 2017-10-10 at 12 30 39 pm



The weekday names should be cropped to the first two letters in the widget view. This is probably better achieved in the markup, though. Something like:
```
<li class="Calendar-cell">
  Mo<span class="u-hiddenVisually">nday</span>
</li>
```
This way the full weekday will still be legible by screenreaders but cropped visually.

captain-pj: Yes! I was going to do just that, but I had hesitations about that for legibility purposes. Is that legal markup in that it won't mess up screen readers and whatnot? This was actually a place where I wanted to discuss a little bit.

src/components/Calendar.scss
captain-pj: Yeah, my intention was just to have this for demo purposes. I can do inline if that is clearer.

src/components/Calendar.scss
captain-pj: Sure- as long as that is okay? Like I said, I didn't want to undo what Bill had already built on this because I wasn't sure if those attributes were needed for something else.

src/components/Calendar.scss
captain-pj: Sure

src/components/Calendar.scss
captain-pj: ? Like I said, some of this comes from what was included in the ticket, but if I can rip it all out, I'm happy to do so!

src/components/Calendar.scss
```diff
@@ -0,0 +1,162 @@
...
+  margin-right: $su-medium;
+  display: inline-block;
+}
+.Calendar-secondary-heading {
+  padding: $su-medium;
+  border-left: 1px solid $ts-grey;
+  border-right: 1px solid $ts-grey;
+  border-top: 1px solid $ts-grey;
+}
+.Calendar-row[data-ui-component="week-heading"] li {
```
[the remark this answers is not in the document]
captain-pj: So some of this is from the existing HTML/CSS in the calendar that Bill included in the ticket. This was how it was set up there. I didn't want to completely rewrite it at first- I wasn't sure if those things were going to need to be there for js hooks or something.

src/components/Calendar.scss
captain-pj: Yeah, I had wondered that too, but wasn't sure how important the mobile-first philosophy was.

src/components/Calendar.scss
captain-pj: Probably not? Was in the original css.

src/components/Calendar.scss
captain-pj: Do we have social media icons? Should I make a ticket for that or just add that to the button component as part of this ticket/branch?

src/components/Calendar.scss
captain-pj: More carryover css. Will clean it up.

src/components/Calendar.scss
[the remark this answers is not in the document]
williamdingwall: @captain-pj  Feel free to re-write that as much as you want.. that original code is pretty old, was playing around with a few different ideas at a time.  So don't feel like you need to keep any of it.

src/components/Calendar.scss
williamdingwall: Feel free to rip it all out and re-write.

src/components/Calendar.scss
williamdingwall: With this, what do you think about using the <abbr> element and replacing with the abbreviated text with the full text at larger screens?

Never a fan of breaking up words with spans.. something along these lines maybe : https://blog.escapecreative.com/how-to-replace-text-with-css/

Then for the content you just pull from the title

saralohr: Seems a little complicated to me... what's the issue with breaking up a word?

saralohr: Hmm.. maybe to move the complication to the CSS rather than the markup?

src/components/Calendar.scss
```diff
@@ -1,127 +1,141 @@
...
-      display: block;
-    }
-  }
-  & > [data-ui-component="current-day"] {
-    background: $cu-primary--bg;
-    span, abbr {
-      color: $cu-primary;
-      font-weight: $tu-bold-fontWeight;
-    }
+.Calendar-cell.Calendar-cell--dayHeading {
```
saralohr: Do these selectors need to be chained?

src/components/Calendar.scss
```diff
@@ -1,127 +1,141 @@
...
-    -webkit-box-flex: 1;
     flex: 1 1 0%;
+    min-height: 0;
+    height: auto;
     box-sizing: inherit;
     min-width: 0;
-    padding: $su-small + 2;
+    text-align: center;
+    padding: $su-small 0;
+    &.Calendar-cell--dayHeading span {
```
saralohr: Do these selectors need to be chained? (`.Calendar-cell.Calendar-cell--dayHeading`)
Same question for classes below...

src/components/Calendar.scss
saralohr: I like this... I think this should probably be classed though, so that if we want a widget at small screens these styles won't apply to it. Something like `Calendar--responsive`. Not sure if we need different versions with different max-breakpoints @williamdingwall

src/components/Calendar.scss
```diff
@@ -1,127 +1,141 @@
...
+
+@media only screen and (max-width: $breakpoint-sm) {
   .Calendar-row {
-    box-sizing: border-box;
-    display: flex;
-    -webkit-box-orient: horizontal;
-    -webkit-box-direction: normal;
-    flex-flow: row wrap;
+    display: block;
+    &.Calendar-row--weekHeading {
```
saralohr: Same chaining question as above... the modifier class _should_ be all you need here.

src/components/Calendar.scss
captain-pj: You mean put them after the responsive stuff? I'm going to move the FB button styles to the button component.

saralohr: Either way... if the responsive stuff gets classed as a modifier then these can both go at the end.

PR #50, opened 2017-12-08
adding collapsable table row modifier

williamdingwall:


src/components/Panel.scss
```diff
@@ -240,3 +241,69 @@ $Panel-header-text-color: $cu-info !default;
...
 }
+
+// * 4. Collapsable Panel Rows
+
+// 1. The child row makes up the bottom border instead of parent row
+// 2. Style is nested for specificity
+// 3. The Parent row needs to be targeted seperately from a static Panel row
+
+.Panel--collapsable {
+  .Panel-parentRow { // 3
```
saralohr: Could this be a modifier instead? ie `Panel-row--parent`

src/components/Panel.scss
```diff
@@ -240,3 +241,69 @@ $Panel-header-text-color: $cu-info !default;
...
+
+.Panel-childRows {
+  background: $cu-middleground;
+  height: 0;
+  overflow: hidden;
+  &:not(:last-child) {
+    border-top: 1px solid $cu-divider--light;
+  }
+  .Panel-row {
+    border: none ;
```
saralohr: Nit: extra space before semicolon.

src/components/Panel.scss
williamdingwall: yep good call

src/components/Panel.scss
williamdingwall: updated

PR #53, opened 2018-01-17
adding tooltip component

williamdingwall:
tooltip inclusion 

screen shot 2018-01-17 at 11 21 36 am

...also made some updates to the states of checkboxes

screen shot 2018-01-17 at 11 24 03 am


src/components/Tooltip.scss
```diff
@@ -0,0 +1,53 @@
...
+// TODO: add position Modifiers
+// TODO: add js function to automatically update position
+
+//------------------------------------------------------------//
+
+// * 1. Base CSS
+
+.Tooltip {
+  position: relative;
+  //cursor: help;
```
saralohr: What do you think about a `cursor: pointer` here? I think that would provide an indication that the content is interactive.

src/components/Tooltip.scss
```diff
@@ -0,0 +1,53 @@
...
+    border-top-left-radius: 0;
+    border: 1px solid $cu-divider--light;
+    background: $cu-middleground;
+    font-size: $tu-base-fontSize;
+    box-shadow: $box-shadow-small;
+    font-style: normal;
+    text-align: left;
+    color: $cu-info;
+  }
+  &:hover:after {
```
saralohr: Could we make this show on `:focus` too? The element would also need a `tabindex="0"` applied if it is not a naturally focusable element.

src/components/index.scss
```diff
@@ -1,14 +1,15 @@
-@import "icon";
-@import "grid";
+@import "Icon";
+@import "Grid";
```
saralohr: Nice! Been wanting to do this for a while. Do these need to be updated in the directory name too?

src/components/index.scss
```diff
@@ -1,14 +1,15 @@
...
 @import "Button";
 @import "ButtonGroup";
 @import "Checkbox";
 @import "Toggle";
 @import "Input";
 @import "SelectBox";
-@import "step-nav";
+@import "StepNav";
```
saralohr: Update directory name?

src/components/Checkbox.scss
```diff
@@ -92,6 +97,12 @@ $Checkbox-bg-color: $cu-foreground !default;
   }
 }
 
+.Checkbox-input:disabled:checked + .Checkbox-label {
+  &:before {
+    background-color: $cu-positive;
+  }
```
saralohr: Just curious - what scenario is this for?

williamdingwall: if the option is selected and can't be unchecked - it's come up a couple of times

src/components/Tooltip.scss
williamdingwall: I went away from that because it's not always on a clickable element, and I've always thought of the pointer as defining click ability specifically. But I'm ok with that, especially with the addition of the focus state option.

src/components/Tooltip.scss
williamdingwall: yep

src/components/index.scss
williamdingwall: yep, they are, surprised that doesn't show in PR

src/components/index.scss
williamdingwall: not sure I follow on this one?  which directory?

saralohr: Oops, I meant CSS file name from step-nav to StepNav

src/components/index.scss
saralohr: On my branch checked out it shows the files as lowercase, still. But StepNav.scss is updated...

PR #55, opened 2018-01-18
first pass on invoicing fit & finish

williamdingwall:
## < Subject > Cleanup for invoicing.

**Story / Task Link**
https://teamsnap.atlassian.net/browse/MNG-3867

### Description
Bringing colors, spacing, and type up to spec with guidelines. Both commissioner and customer facing sides. Crosses Classic, League Frontend, and TeamSnap-UI

### What did I do?
- Update Gird component for alignment with gutter modifier
- update input and checkbox for consistency with classic
- adding placeholder to base for consistency with classic

### Related PRs
Classic : https://github.com/teamsnap/classic/pull/7650
LFE : https://github.com/teamsnap/league-frontend/pull/418
TeamSnap-UI : https://github.com/teamsnap/teamsnap-ui/pull/55


src/base/reset.scss
```diff
@@ -20,3 +20,23 @@ img {
   height:    auto;
   border:    none;
 }
+
+::-webkit-input-placeholder {
+  color:     scaleColor($ts-black, 6);
+  font-size: $tu-small-fontSize;
```
saralohr: What is the default font-size? Does it not inherit?

src/base/reset.scss
```diff
@@ -20,3 +20,23 @@ img {
...
+}
+
+:-moz-placeholder { /* Firefox 18- */
+   color: scaleColor($ts-black, 6);
```
saralohr: Nit: indent

src/components/Grid.scss
```diff
@@ -60,12 +60,14 @@ $Grid-gutter: $su-medium;
...
-}
-
 .Grid--withGutter > .Grid-cell {
   padding: 0 (0.5 * $Grid-gutter);
+  &:first-child {
+    padding-left: 0;
+  }
+  &:last-child {
+    padding-right: 0;
+  }
```
saralohr: Hrm.. how does this work? It's not just the first and last child we want to target.. right? 

I also have reservations about straying from the proven SUIT Grid pattern. The only reason it's broken is because of us adding this: https://github.com/teamsnap/teamsnap-ui/pull/55/commits/7c5a9c9fe627468415e8de4afc2f44573ae33bea#diff-f238cb697777bedff815cc431443ab2fR22

I think the better thing to do would be to remove that and address the instances where we need the Grid to be 100%.

src/base/reset.scss
williamdingwall: oh that was just copied from classic - I'm cool with letting it inherit

src/base/reset.scss
williamdingwall: got it

src/components/Grid.scss
williamdingwall: It wouldn't accommodate for grid cells dropping to new lines but assuming it's always..

```
<div class="Grid Grid--gridWithGutter">
  <div class="Grid-cell"></div>
  <div class="Grid-cell"></div>
  <div class="Grid-cell"></div>
</div>
```
..works fine

Didn't realize it was something added to LFE breaking it - yeah I'm cool with removing that width, and seeing what that might break

saralohr: Okay - I can take that on. We added it fix some of the full-width tables, but we'll have to find a work-around.

williamdingwall: Oh I see on the checkout, that wouldn't work for both checkout and overview usage.  Added a new width declaration with CSS calc() that seems to do the trick for both. Looks like it should cover our basis too : https://caniuse.com/#feat=calc

```
.Grid--withGutter {
  margin: 0 (-0.5 * $Grid-gutter);
  width: calc(100% + 16px); // 2
}
```

thoughts on that?

src/components/Grid.scss
```diff
@@ -62,6 +64,7 @@ $Grid-gutter: $su-medium;
 
 .Grid--withGutter {
   margin: 0 (-0.5 * $Grid-gutter);
+  width: calc(100% + 16px); // 2
```
saralohr: I'm cool with this approach.

Re: variable: I think this might work with interpolation: `#{$Grid-gutter}`... also I think Sass has operators prebaked, so you may not need calc at all, like in the above margin expression. http://sass-lang.com/guide#topic-8

williamdingwall: 1. ah that's a good call on the #{$Grid-gutter}. Updated.

2. That's definitely an interesting approach! and works well for specific scenarios because we know the max width of a given element (full container for example is 1300px). So `width: ($max-width + $Grid-gutter) / $max-width  * 100%;` covers the overview screen but it doesn't seem to work for the checkout page because we don't know what that container width is in terms of a pixel value, and may not know what it is in future situations.  In the example too they're working off set values to create those percentages, just don't _think_ we have that luxury here. Whereas calc(100%) always returns what that container width ends up being in the browser and let's us add on to it.

saralohr: Cool 👍 Yeah, I was thinking of just using 100% + $Grid-gutter, but I didn't realize the operators don't function quite like `calc`, in that they _pre_ calculate versus calculate on render.

src/components/Grid.scss
```diff
@@ -62,6 +62,7 @@ $Grid-gutter: $su-medium;
 
 .Grid--withGutter {
   margin: 0 (-0.5 * $Grid-gutter);
+  width: 1316 / 1300 * 100%;
```
saralohr: Wait... why not use calc(100% + #{$Grid-gutter})?

PR #56, opened 2018-02-14
Add themes and initial gulp tasks

saralohr:
This is a first pass at prepping TeamSnap-UI for publishing CSS to CDN.

What I did:
+ Added color themes from Classic
+ Set up a bunch of Gulp tasks to build our CSS
+ Updated font paths (I thought maybe we could publish fonts from here after all... the app will only download the font the first time it gets the request I think... otherwise it will use the one it already has - need to check this)

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,9 @@
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
     "lint": "sass-lint 'src/**/*.scss' -v",
-    "prepublish": "npm run lint"
+    "prepublish": "npm run lint",
+    "prestart": "npm install",
+    "start": "gulp"
   },
   "repository": {
     "type": "git",
@@ -24,6 +26,14 @@
     "suit"
   ],
   "devDependencies": {
+    "browser-sync": "^2.23.6",
+    "del": "^3.0.0",
+    "gulp": "^3.9.1",
+    "gulp-autoprefixer": "^4.1.0",
+    "gulp-cssnano": "^2.1.2",
+    "gulp-if": "^2.0.2",
+    "gulp-sass": "^3.1.0",
+    "gulp-util": "^3.0.8",
     "sass-lint": "^1.10.2"
   }
 }
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "test": "echo \"Error: no test specified\" && exit 1",
     "lint": "sass-lint 'src/**/*.scss' -v",
     "prepublish": "npm run lint",
-    "prestart": "npm install",
+    "install": "gulp build",
     "start": "gulp"
   },
   "repository": {
--- a/src/fonts/museo.scss
+++ b/src/fonts/museo.scss
@@ -6,60 +6,60 @@
 
 @font-face {
   font-family: 'MuseoSansRounded100Regular';
-  src: url('../fonts/museo/MuseoSansRounded-100-webfont.eot');
-  src: url('../fonts/museo/MuseoSansRounded-100-webfont.woff') format('woff'),
-    url('../fonts/museo/MuseoSansRounded-100-webfont.ttf') format('truetype'),
-    url('../fonts/museo/MuseoSansRounded-100-webfont.svg') format('svg');
+  src: url('/assets/fonts/museo/MuseoSansRounded-100-webfont.eot');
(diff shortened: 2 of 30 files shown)


mikereinmiller:
Did we ever decide how we wanted to handle the inline SVG icons?

saralohr:
> Did we ever decide how we wanted to handle the inline SVG icons?

Not yet - It's still an open question, but for now people can continue to access them from the node module.

package.json
```diff
@@ -6,7 +6,9 @@
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
     "lint": "sass-lint 'src/**/*.scss' -v",
-    "prepublish": "npm run lint"
+    "prepublish": "npm run lint",
```
Ollo: We could auto build on if we want by updating this to : `npm run lint && npm run <gulp  build task here>` 
Not sure which gulp task we should map to an npm script but this would ensure we have a fresh compile in `/dist` before its published to npm for usage.

matthopson: @Ollo Will that matter if we aren't consuming it directly from the package? Seems like that step might bet better on post install, so you're not lugging the `dist` folder around. Either way, good call to automate that step (whenever the appropriate time is).

What are your thoughts on deployment to CDN?. If we're versioning, the build server could run the build, read package.json for the version number and then deploy to a versioned folder name on dugout, yeah?

Ollo: Since publishing has typically been the only spot where changes are consistently versioned I was thinking that might be the appropriate place to build the `dist` folder and copy it to the cdn in a versioned directory structure `dugout.teamsnap.com/teamsnap-ui/v2.4.0/`. ?

matthopson: For reference, when I think "publish" I think `npm publish`.... when I think deploy, I think `jarvis deploy...`.

I don't necessarily see a ton wrong with building `dist` on publish in that case (we do it with the SDK), but I'm curious if there will be env considerations re: deploy, that would cause the build to need to re-run, anyway.

saralohr: @Ollo @matthopson Here's what I ended up doing with scripts:

+ For deployment to CDN, we'll be telling Ops which task to run on `jarvis deploy`. I'm thinking `gulp build` is the right one, since it cleans out the `dist` directory and runs through all the build steps, without starting any servers or watching anything. 

+ Ops also said they could reference the version from the package.json for creating the directory structure on the CDN when we deploy. Awesome. Great idea @matthopson! 

+ Left `prepublish` untouched since that seems good as is. 

+ Added `gulp build` to `npm install` which will add all the `dist` goodness when you install the node module, should anyone want to reference or pull from these assets (don't think this will be a common use case, but figured it doesn't hurt)

+ `npm start` runs `gulp`, which in addition to `build`ing, starts a server and watches files. This is for local development from the repository.

src/fonts/museo.scss
```diff
@@ -6,60 +6,60 @@
 
 @font-face {
   font-family: 'MuseoSansRounded100Regular';
-  src: url('../fonts/museo/MuseoSansRounded-100-webfont.eot');
-  src: url('../fonts/museo/MuseoSansRounded-100-webfont.woff') format('woff'),
-    url('../fonts/museo/MuseoSansRounded-100-webfont.ttf') format('truetype'),
-    url('../fonts/museo/MuseoSansRounded-100-webfont.svg') format('svg');
+  src: url('/assets/fonts/museo/MuseoSansRounded-100-webfont.eot');
+  src: url('/assets/fonts/museo/MuseoSansRounded-100-webfont.woff') format('woff'),
```
saralohr: I still don't know what the frack I'm doing with fonts. I think these need something other than a relative path to work in production, locally, and referenced from different apps, so we either can use the assets from Classic or reference what will be published to CDN from here. I'm deferring this question for now and figure we can experiment once these are up on CDN.

package.json
```diff
@@ -6,7 +6,9 @@
...
+    "install": "gulp build",
```
Ollo: I think using `postinstall` is a better method for compiling locally after install. My only worry with overloading install is it could produce unintended side effects when running this through various CI platforms?

saralohr: I'm fine with `postinstall`. Not sure the implication of CI. If we run into issues it could be removed from the task altogether.

saralohr: It looks like `install` and `postinstall` both run after the package is installed: https://docs.npmjs.com/misc/scripts

PR #57, opened 2018-02-19
Update form components

saralohr:
+ Refactor `Input` and add `FormGroup` and `InputGroup` for more flexible form handling
+ Include `Checkbox` and `SelectBox` in form error styles
+ Add additional spacing classes

### DIFF
--- a/src/components/Input.scss
+++ b/src/components/Input.scss
@@ -5,44 +5,28 @@
 // * Input
 // * 1. Input Settings
 // * 2. Base Input
-// * 3. Inline Modifiers
-// * 4. Amount and Calendar Modifier
-// * 5. Errors and feedback
 
 // Notes:
-// * size utilities should be used on the Input-element to
-// modify sizing
 // * all the classic element styling is coming through
 
 // TODO : Refactor classic form styling to resolve conflicts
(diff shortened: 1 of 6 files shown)


src/components/Input.scss
```diff
@@ -25,23 +19,14 @@ $form-input-height: $su-xlarge !default;
...
-  font-size: $tu-small-fontSize;
-  font-weight: $tu-semibold-fontWeight;
-}
-
-.Input-element {
   cursor: text;
-  height: $form-input-height;
+  display: block;
+  width: 100%;
+  height: $form-input-height !important; // a.
```
williamdingwall: I'm wondering if we should just update that static stylesheet or link directly to the one in classic instead of adding things like !important to the new CSS.  There were some changes in classic that should allow you to override with just a class

williamdingwall: here's what they're doing in LFE https://github.com/teamsnap/league-frontend/blob/5185b0925544976a12bb1f6849c48a1e58620d04/src/index.html#L7

saralohr: Oh, hmmm. That's a good point that we could just update the style in classic that is setting the height too small, since we want it to be that way in Classic too.

PR #59, opened 2018-03-15
Adds React components

Ollo:
#### UI Patterns
https://teamsnap-ui-patterns.netlify.com/

#### Details
- Updates directory structure to provide assets css and js 
- Adds tooling to support js ( babel )
- Adds base patterns for components / props based on patterns library
- Adds helper method for combining className
- Bumps version to 2.0.0

#### TODO - v1
- [x] Add Expandable Panel 'Table' Rows
- [x] Add Component Documentation
- [x] Updates README / Docs
- [x] Decide on starting JS linter (Airbnb, Standard, Custom)?

#### TODO - v2
- Add StepNav 
- Add SummaryList
- Add Radial Progress Bar
- Add PanelWrapper for easier use of Panel defaults (similar to FieldWrapper)
- Add Table for easier use of Panel table defaults (similar to PanelRowExpanded)

*Adds the following simple UI components*
```
  Button,
  ButtonGroup,
  Cell,
  Checkbox,
  Divider,
  FieldGroup,
  FieldLabel,
  FieldMessage,
  FieldWrapper,
  Grid,
  Icon,
  Input,
  InputControl,
  Loader,
  Panel,
  PanelBody,
  PanelCell,
  PanelFooter,
  PanelHeader,
  PanelRow,
  PanelRowExpandable,
  ProgressBar,
  Radio,
  Select,
  TextArea,
  TextLink,
  Toggle,
  Tooltip
```


src/js/components/Button/Button.js
```diff
@@ -0,0 +1,99 @@
...
+    const { children, type, style, isDisabled, onClick } = this.props
+
+    return (
+      <button
+        type={ type }
+        className={ this.getButtonClassName() }
+        style={ style }
+        onClick={ onClick }
+        disabled={ isDisabled }
+        children={ children } />
```
amberwebb: Any reason to have children as a prop here? It makes sense for a link that could hold a node, but a button will only hold text.

src/js/components/Button/Button.js
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  render () {
+    const { location } = this.props
+    return location ? this.renderButtonLink() : this.renderButton()
+  }
+}
+
+Button.propTypes = {
+  children: PropTypes.node.isRequired,
```
amberwebb: I'm not sure this should always be required, since a button will not have children per se. It will have text. Unless I'm thinking of this wrong. An href could hold a node or text.

.sass-lint.yml
```diff
@@ -31,25 +29,36 @@ rules:
...
+        - height
+        - color
+        - background-color
+        - font-size
+        - letter-spacing
+        - line-height
+        - font-family
+        - border-radius
+        - border-color
+        - border-width
```
dmccraw: eof

src/js/components/Cell/index.js
```diff
@@ -0,0 +1,2 @@
+import Cell from './Cell'
+export default Cell
```
dmccraw: eof

src/js/components/Checkbox/index.js
```diff
@@ -0,0 +1,2 @@
+import Checkbox from './Checkbox'
+export default Checkbox
```
dmccraw: eof

src/js/components/Checkbox/Checkbox.js
```diff
@@ -0,0 +1,29 @@
...
+}
+
+Checkbox.propTypes = {
+  name: PropTypes.string.isRequired,
+  label: PropTypes.string.isRequired,
+  inputProps: PropTypes.object,
+  isInline: PropTypes.bool,
+  className: PropTypes.string,
+  mods: PropTypes.string,
+  style: PropTypes.object
```
amberwebb: `mods` are modifiers? What's the difference between adding additional class names to `className` and adding something to `mods`? I don't quite understand that.

src/js/components/FieldGroup/FieldGroup.js
```diff
@@ -0,0 +1,56 @@
...
+class FieldGroup extends PureComponent {
+  renderFieldComponent = () => {
+    const { name, field, fieldProps } = this.props
+
+    let FieldTag = Input
+    if (field === 'select') { FieldTag = Select }
+    else if (field === 'checkbox') { FieldTag = Checkbox }
+    else if (field === 'radio') { FieldTag = Radio }
+    else if (field === 'toggle') { FieldTag = Toggle }
+    else if (field === 'textarea') { FieldTag = TextArea }
```
dmccraw: Feels like this could be a map
```
const FieldTypes = {
  select: Select,
  checkbox: CheckBox,
  radio: Radio,
  toggle: Toggle,
  textarea: TextArea
}
const FieldTag = (FieldTypes[field] || Input)
```
Makes is super easy to add field types

src/js/components/PanelBody/PanelBody.js
```diff
@@ -0,0 +1,31 @@
...
+
+    return (
+      <div className={ getClassName(className, mods) } style={ style }>
+        { children }
+      </div>
+    )
+  }
+}
+
+
```
dmccraw: style nit: There are some places with 2 newlines and some without? Any particular reason?

dmccraw:
No javascript linter?

dmccraw:
This looks fantastic! Amazing you got so much done in a week. I would love to see a linter and some of the eof's fixed but great job!

src/js/components/Input/Input.js
```diff
@@ -0,0 +1,45 @@
+import React, { PureComponent } from 'react'
+import PropTypes from 'prop-types'
+import { getClassName } from '../../utils/helpers'
+
+class Input extends PureComponent {
```
saralohr: Should this be `InputGroup` instead.. since it's outputting that component? 
http://localhost:3002/patterns/components/input-group.html

And `Input` would just be the base `.Input`? 
http://localhost:3002/patterns/components/Input.html

src/js/components/FieldMessage/FieldMessage.js
```diff
@@ -0,0 +1,41 @@
+import React, { PureComponent } from 'react'
+import PropTypes from 'prop-types'
+import { getClassName } from '../../utils/helpers'
+
+class FieldMessage extends PureComponent {
+  render () {
+    const { children, isError, className, mods, style } = this.props
+
+    // TODO: Can we adjust this classStructure to be a modifier instead of replacement?
+    //  - FormGroup-message & FormGroup-feedback are almost identical except for color
```
saralohr: I think we kept these separate in case there were _both_ : a `.FormGroup-message` that is instructional and always shows, vs a `.FormGroup-feedback` which provides feedback based on form validation, and thus should be styled independently of the message. I suppose we could replace the instructional text with the feedback text.. and supply styles based on the `.is-invalid` class of the `FormGroup` as you suggest. Thoughts @williamdingwall?

src/js/components/Field/Field.js
```diff
@@ -0,0 +1,46 @@
...
+  children: PropTypes.node.isRequired,
+  status: PropTypes.oneOf([null, 'success', 'error']),
+  className: PropTypes.string,
+  mods: PropTypes.string,
+  style: PropTypes.object
+}
+
+Field.defaultProps = {
+  status: null,
+  className: 'FormGroup',
```
saralohr: I'm not sure what the solution is.. but this throws me off: `FormGroup` is the CSS pattern class used for `Field` but then there is a separate `FieldGroup` component that is pulling in `Field`. Could `Field` and `FieldGroup` be combined maybe to just `FieldGroup`? Or `FormGroup` even? I'm open to changing the CSS component to `FieldGroup` alternately... or are you intentionally keeping the naming convention different since the component structure isn't parallel to the CSS pattern structure?

src/js/components/Icon/Icon.js
```diff
@@ -0,0 +1,36 @@
+import React, { PureComponent } from 'react'
+import PropTypes from 'prop-types'
+
+const svgIcon = (name) => require(`../../icons/${name}`)
+
+// TODO: Icons flash 'ginormous' on page load, before css kicks in and adjust
```
saralohr: A solution to this is to set `width` and `height` attributes on the source SVG, preventing them from filling the screen until the CSS loads, and once it does the `.Icon` class overrides these properties. I thought I had includes those attributes already, but I guess not! I can get those attributes added.

src/js/components/Field/Field.js
```diff
@@ -0,0 +1,46 @@
...
+      <div className={ fieldClasses } style={ style }>
+        { status === 'success' && this.renderSuccess() }
+        { children }
+      </div>
+    )
+  }
+}
+
+Field.propTypes = {
+  children: PropTypes.node.isRequired,
```
saralohr: Would this support multiple children? For example in the case of multiple checkboxes or radios that have validation that applies to them as a group (ie they are required). Example that shows the `FormGroup` pattern with two nested radio buttons that are required:

form-group-multiple-children

src/js/components/PanelCell/PanelCell.js
```diff
@@ -0,0 +1,39 @@
...
+
+class PanelCell extends PureComponent {
+  renderChildrenHeader = (children) => <h4 className="Panel-title">{ children }</h4>
+
+  render () {
+    const { children, isHeader, className, mods, style } = this.props
+
+    // Thoughts: The docs had to variations of a heading, can we update this or are they different use cases?
+    //  * Panel-cell & Panel-cell--header
+    //  * Panel-cell > Panel-title
```
saralohr: Oh.. yeah, this is actually two different things. `Panel-title` is for the main `Panel` title in the main `Panel-header`, the other (`Panel-cell--header`)is when a Panel cell is acting as the label or title for adjacent content:


panel

src/js/components/PanelCell/PanelCell.js
```diff
@@ -0,0 +1,39 @@
+import React, { PureComponent } from 'react'
+import PropTypes from 'prop-types'
+import { getClassName } from '../../utils/helpers'
+
+class PanelCell extends PureComponent {
+  renderChildrenHeader = (children) => <h4 className="Panel-title">{ children }</h4>
```
saralohr: So in this case the parent `Panel-cell--header` is applying style to the text element, so `Panel-title` isn't needed.

src/js/components/Select/Select.js
```diff
@@ -0,0 +1,45 @@
...
+    const { label, value, disabled } = option
+
+    return <option key={ value } value={ value } disabled={ disabled }>{ label }</option>
+  }
+
+  render () {
+    const { name, options, className, mods, style, inputProps } = this.props
+
+    return (
+      <label className={ getClassName(className,mods) } style={ style } htmlFor={ name }>
```
saralohr: I think this can actually be just a `<div>` wrapper instead of a `label` so that we can use the standard `FormGroup-label` as a sibling element. (I know it's written in the pattern library as a `label` but I think it will be more flexible this way - I can update the markup example)

saralohr:
🙌 You guys rocked this! 

I left some CSS-related questions/comments, but this is awesome! Super impressive that ya'll knocked this out in a week! 👏 💥 💯

src/js/components/Button/Button.js
mikereinmiller: So a button could also hold an Icon.  So we could allow the button to accept `text` and `icon` and even an `iconPosition`, but doing something like children gives full flexibility.  I'm open to a change here though if we would rather.

src/js/components/Button/Button.js
mikereinmiller: If button only accepts children as say the text or icon, then I think it's safe to have required here.

src/js/components/Cell/index.js
mikereinmiller: Adding a specified linter with the spacing would fix your concerns here.  i don't think it's a big deal.  🤷‍♂️   But we can agree on a linter say standard, airbnb or another.

src/js/components/Checkbox/index.js
mikereinmiller: Adding a specified linter with the spacing would fix your concerns here.  i don't think it's a big deal.  🤷‍♂️   But we can agree on a linter say standard, airbnb or another.

src/js/components/Checkbox/Checkbox.js
mikereinmiller: So in this case it helps save you the understanding of what the base class needs to be.  So for an example of say button, the default class is 'Button'.  The following two will allow for the same outputs, except in one case you don't need to rewrite or even know about the base class.  `<Button className='Button someUniqueModifier' />` and `<Button mods='someUniqueModifier' />`

src/js/components/FieldGroup/FieldGroup.js
mikereinmiller: I'm not opposed to a change like this.  I don't really see this changing as there are not a ton of 'input' field types.  But it seems to be a nice way to clean it up.  👍

src/js/components/PanelBody/PanelBody.js
mikereinmiller: Adding a specified linter with the spacing would fix your concerns here.  i don't think it's a big deal.  🤷‍♂️   But we can agree on a linter say standard, airbnb or another.

src/js/components/Input/Input.js
mikereinmiller: I went back and forth on if I should split `Input` from `InputGroup`.  Through my testing, I saw no impact on always allowing the div wrapper of InputGroup.  So I thought this might simplify the approach when trying to add the appropriate icon, instead of trying to remember when.  I'm open to a change here as well.

src/js/components/FieldMessage/FieldMessage.js
mikereinmiller: That's a good thought.  I wouldn't see both the message and error showing at the same time, I guess I thought them as replacing when needed.  But I could be wrong, either way I suppose.

mikereinmiller: That's a good thought.  I wouldn't see both the message and error showing at the same time, I guess I thought them as replacing when needed.  But I could be wrong, either way I suppose.

src/js/components/Field/Field.js
mikereinmiller: So agree this could look a little confusing and here is where the thought came from.  😬 

The idea of FormGroup, I found very confusing since this is pretty much just a Field 'wrapper' around various inputs that includes the appropriate spacing, labels and messages.  FormGroup to me sounds like a grouping of Form data or multiple fields.

The idea with FieldGroup (currently) is a react helper that wraps the components needed for a field but it has some limitations (currently) that only allow the default settings.  if you want more advances settings, styles, etc you would want to craft the appropriate Field components on your own.

src/js/components/Grid/Grid.js
```diff
@@ -0,0 +1,47 @@
...
+class Grid extends PureComponent {
+  render () {
+    const { children, isFit, isEqualHeight, isAlignCenter, isAlignMiddle, className, mods, style } = this.props
+
+    const gridClasses = getClassName(
+      className,
+      isFit && 'Grid--fit',
+      isEqualHeight && 'Grid--equalHeight',
+      isAlignCenter && 'Grid--alignCenter',
+      isAlignMiddle && 'Grid--alignMiddle',
```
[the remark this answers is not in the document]
mikereinmiller: oh good catch, looks like I missed one.  We should get that added.  Also remember at anytime, you can pass the className or even various modifiers via the className or the mods attribute.  These should be useful when needing to chain various utility methods

src/js/components/Icon/Icon.js
mikereinmiller: Cool, if I messed something up in the component or need to add styles, just let me know.,

src/js/components/Icon/Icon.js
```diff
@@ -0,0 +1,36 @@
...
+  size: PropTypes.string,
+  color: PropTypes.string,
+  className: PropTypes.string,
+  style: PropTypes.object
+}
+
+Icon.defaultProps = {
+  size: 'inherit',
+  color: 'inherit',
+  className: 'Icon',
```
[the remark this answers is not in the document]
mikereinmiller: 🤔 I see no issues with that.  Lets talk about this more on Monday.  I think there is something there that can help simplify and clean this up.  👍

src/js/components/Field/Field.js
mikereinmiller: This should work like you said, you could pass 2 checkboxes say inline that would render right next to each other under a single label and with a single set of errors.

src/js/components/PanelCell/PanelCell.js
mikereinmiller: hmm...ok. I'll take a closer look at this and make the appropriate changes.  I'll hit you up if I have any questions.

src/js/components/Select/Select.js
mikereinmiller: i can change this as well.  Does that mess up the 'htmlFor' we are using to select the inputs/selects?  We can chat more on monday as well.

src/js/components/Button/Button.js
dmccraw:
```
  children: PropTypes.oneOfType([
    PropTypes.arrayOf(PropTypes.node),
    PropTypes.node,
  ]),
```
What if you want to send in an array of nodes?
```
<Button>
 <icon>
 Text
</Button>
```

src/js/components/Button/Button.js
dmccraw: I have seen children as a nested component such as
```
<button> 
  {children}
</button>
```
Does it make any difference passed as a prop?

src/js/components/Button/Button.js
mikereinmiller: I don't really think that is necessary.  I think just accepting children should be enough for these buttons, right?

src/js/components/Button/Button.js
mikereinmiller: no, I think either one would be accepted as the same 'prop'

src/js/components/Cell/index.js
dmccraw: The problem is if anyone else automatically adds newlines to the end of files then anytime they view this file it will get updated even though they just viewed the file. I would like to set the standard that all files have a newline at the end.
https://robots.thoughtbot.com/no-newline-at-end-of-file

src/js/components/Field/Field.js
saralohr: I'm cool with the name `FieldGroup` instead.. `FormGroup` is what Bootstrap uses for this pattern so I just stole that name for lack of a better idea. I'd like to update the CSS to use this same name to prevent confusion.

Interesting distinction between `Field` and `FieldGroup`... I get it though. I guess the reality is that there won't always be a strict 1:1 comparison of React components to CSS patterns.

mikereinmiller: yeah, maybe so.  I think we will be pretty close on the core classes though.  We should find some time and talk through these a bit.

src/js/components/FieldMessage/FieldMessage.js
saralohr: > I wouldn't see both the message and error showing at the same time, I guess I thought them as replacing when needed.

Does this get complicated to save and swap messages vs feedback for each field? For example, if message text is replaced by error text and then needs to go back to the message text once the field is valid? Alternately we could keep the separate elements (`-message` and `-feedback`) and hide and show each based on parent state.

> Can we adjust this classStructure to be a modifier instead of replacement?
Not sure what you mean by this.

mikereinmiller: Nothing super complicated either way.   I think it's more of a design opinion on how you guys think it should operate.  Do you want to see the information message at all times even if an error is created or should the error temporarily replace the information message when triggered?

So if error and message are essentially the same field wouldn't it make more sense to make the class structure something like `className="FormGroup-message FormGroup-message--error"`

src/js/components/Grid/Grid.js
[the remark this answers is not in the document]
saralohr: > you can pass the className or even various modifiers via the className or the mods attribute

Love this!

src/js/components/Select/Select.js
saralohr: No, it shouldn't. We'd just put the `htmlFor` on the separate `-label` instead.

Ollo: Just ran a quick experiment and React won't handle the focus behavior event if you move the htmlFor to a div and not a sibling label. https://codesandbox.io/s/o9n1xm07zz

saralohr: I wasn't suggesting `htmlFor` goes on a div... but that it goes on a sibling `<label>` element rather than a wrapping `<label>` element. Are you saying React won't handle focus if the form element isn't wrapped in a `<label>`? That's crazy...

src/js/components/FieldMessage/FieldMessage.js
saralohr: Ah, I see what you mean now. Yes, using a modifier does make more sense.

src/js/components/FieldGroup/FieldGroup.js
```diff
@@ -0,0 +1,60 @@
+/**
+ * @name FieldGroup
+ * 
+ * @description
+ *  A field group component is a wrapper for grouping the input with label and messages.  See the teamsnap patterns
+ *  library for more information. https://teamsnap-ui-patterns.netlify.com/patterns/components/form-group.html
```
saralohr: TODO: Update this URL once UI-Patterns is updated with FieldGroup.

src/js/components/FieldWrapper/FieldWrapper.js
```diff
@@ -0,0 +1,76 @@
+/**
+ * @name FieldWrapper
```
saralohr: I like this 👍

src/js/components/Select/Select.js
Ollo: No it just has to be a `label` in order for it to properly pass the `for` attribute. 👍

PR #60, opened 2018-03-21
Add color and font size utilities

saralohr:


src/utils/text.scss
```diff
@@ -41,3 +41,32 @@
...
+
+.u-fontSizeXs {
+  font-size: $tu-xsmall-fontSize !important;
+}
+
+.u-fontSizeSm {
+  font-size: $tu-small-fontSize !important;
+}
+
+.u-fontSizeBase {
```
saralohr: I kept this `base` vs `medium` as we did with spacing to preserve it as the base of our modular scale. I'm open to something else though if you disagree @williamdingwall.

williamdingwall: 'Md' might make more sense with everything else.. seems like with 'Base' part of the class we're trying to describe usage too much instead of just providing a consistent scale of classes to work with

mikereinmiller:
I like the idea of these color options, curious if they would change any of the current modifiers?  For example: `Button Button--primary`?

saralohr:
I responded to Mike already, but for posterity adding the answer here:

> I like the idea of these color options, curious if they would change any of the current modifiers? For example: Button Button--primary?

These are just changing `color`. Most of our other modifiers are changing more properties than that. For the button example, background color, border, box shadow and color all change.

So these color utilities allow us more granular control to override a default color when necessary.

PR #65, opened 2018-03-23
StepNav Component

mikereinmiller:
This pull request includes some small bug-fixes as well as adds the StepNav component.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "2.0.3",
   "description": "a CSS component library for TeamSnap",
   "main": "dist/js/index.js",
+  "sideEffects": false,
   "scripts": {
     "lint:js": "eslint src/js/ || true",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -3,6 +3,7 @@
   "version": "2.0.3",
   "description": "a CSS component library for TeamSnap",
   "main": "dist/js/index.js",
+  "sideEffects": false,
```
Ollo: What does this address?

mikereinmiller: I was doing some treeshaking testing with Webpack 4 (forgot to remove from this branch).  Since we are currently still building with babel, this can be removed.  Good catch.   

In general in case you were curious - https://medium.com/webpack/road-to-webpack-4-week-20-21-1641d03ce06e

mikereinmiller:
deleting this in favor of the SummaryList branch (included StepNav)

PR #66, opened 2018-03-23
DES-56 - Add Modal CSS

saralohr:
For markup and demo, see UI-Patterns PR: https://github.com/teamsnap/ui-patterns/pull/44

screen shot 2018-03-23 at 1 53 13 pm
screen shot 2018-03-23 at 1 53 23 pm

### DIFF
--- a/src/css/components/Modal.scss
+++ b/src/css/components/Modal.scss
@@ -1,0 +1,82 @@
+$Modal-bgColor: $cu-foreground !default;
+$Modal-overlayColor: $ts-black !default;
+$Modal-borderColor: $cu-background !default;
+$Modal-borderWidth: $su-small;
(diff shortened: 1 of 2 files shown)


src/css/components/Modal.scss
```diff
@@ -0,0 +1,82 @@
...
+  align-items: center; // 2
+}
+
+.Modal.is-open {
+  display: flex; // 2
+}
+
+.Modal.is-closing {
+  animation: overlayOut $Modal-animation forwards;
+}
```
saralohr: Depending on how we handle the JS for this pattern, some of this may need to change. For now these states exist to show the modal and to animate its closing.

mikereinmiller:
This looks pretty good.  Looks like the core 'functionality' of the modal lives by toggling a set of classes on the root modal div?

PR #68, opened 2018-03-30
Add grey color utility and negative margin utility

saralohr:
Adds utilities to support table header controls:
- grey color utility for $ts-grey @williamdingwall foreground/background/middleground were all too light or too dark. $ts-grey gets themed, right?
- negative margin utility for right-aligned sorting arrows
- Examples in UI-Patterns PR: https://github.com/teamsnap/ui-patterns/pull/45

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@teamsnap/teamsnap-ui",
-  "version": "2.1.0",
+  "version": "2.1.1",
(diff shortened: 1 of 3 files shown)


saralohr:
> Do we have any plans or thoughts to add the other color "scales" to the utilities?

There is so many of them available... we can add them on an as-needed basis I think. So just shout anytime you need one or anticipate needing one and we can pop it in there.

PR #69, opened 2018-04-02
Table component

mikereinmiller:
#### Summary
This PR adds the Initial implementation for the Table Component.

Even though this table component is still a work in progress, it is at a state that it can be merged and we can build upon it with additional rolling features (listed below).  That way we can start to use this table component and know that it will just be updating the props to render additional table options

#### Changes
- [x] Add basic Table styling using Panels
- [x] Accept arrayOf objects of columns and rows for organizing data
- [x] Add Table Sorting with custom sort option
- [x] Allow Table cells to use custom rendering
- [ ] Add Pagination Component
- [ ] Add ability to display checkbox columns for selecting rows (right or left), should also provide a rowSelect callback handler.
- [ ] Table should be able to handle different states with Overrides such as A Placeholder, A Loading State and a No Results found after search
- [ ] Allow ability to pass in additional buttons next to the pagination, could also just look at rendering a custom header and footer components
- [ ] Add TableFiltering Component
- [ ] Implement Expandable Rows (deprecate PanelRowExpandable?)

### DIFF
--- a/src/js/components/Table/Table.js
+++ b/src/js/components/Table/Table.js
@@ -1,0 +1,138 @@
+import React, { PureComponent } from 'react'
+import PropTypes from 'prop-types'
+import { setUniqueId } from '../../utils/helpers'
+import { sortBy } from '../../utils/sort'
+import Icon from '../Icon'
+import Panel from '../Panel'
+import PanelBody from '../PanelBody'
+import PanelRow from '../PanelRow'
+import PanelCell from '../PanelCell'
+import TextLink from '../TextLink'
+
+class Table extends PureComponent {
+  state = {
+    items: [],
+    sortByColumn: null,
+    sortByDirection: false
+  }
+
+  componentWillMount() {
+    const { rows, defaultSort } = this.props
+    const items = setUniqueId(rows)
+    const tableState = defaultSort ? this.sortItems(defaultSort, items) : { items }
+
+    this.setState({ ...tableState })
+  }
+
+  sortItems = (newSort, newItems) => {
+    const { columns } = this.props
+
+    // Check if newSort has a `-`, representing reverse sorting
+    let sortByReverse = newSort.charAt(0) === '-' ? true : false
+    const sortByColumn = sortByReverse ? newSort.substr(1) : newSort
+
+    // If newSort has same name as previous sort - reverse sort direction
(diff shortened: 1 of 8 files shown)


src/js/components/Table/Table.js
```diff
@@ -0,0 +1,175 @@
...
+    return (
+      <TextLink location={ name } onClick={ this.handleSortClick } mods="u-flex">
+        { this.renderSortLabel(label) }
+        <div className="u-colorGrey u-fontSizeXs u-spaceLeftXs">
+          <Icon name="up" mods={ `u-block ${ascLinkClassName}` } />
+          <Icon name="down" mods={ `u-block ${descLinkClassName}` } />
+        </div>
+      </TextLink>
+    )
+  }
```
saralohr: This is cool! 

Do we want to build in a right-align option? This is currently used for dollar value columns in LFE (also mocked in the complex table demo in UI Patterns) It basically just adds a couple more classes to the link wrapper: `u-flexJustifyEnd u-spaceNegativeRightSm`. And then the column content cells just get `u-textRight` on each of them. Or do you think these should just get added as additional `mods` when the component is used?

mikereinmiller: That's a good question.  We could set an `alignment` attribute on the columns that could maybe render: right, left, or center.  This would set the appropriate utility methods like `u-textRight or the parent and `u-flexJustifyEnd u-spaceNegativeRightSm` on the link?  W could also automatically try and add the left/right/center justification by checking if the 'u-textRight/Left' mod was passed?

saralohr: Hmm, I'd be fine either way, whatever you think is easier for JS devs. An `alignment` attribute _might_ be easier?

src/js/components/Table/Table.js
```diff
@@ -0,0 +1,187 @@
...
+
+  render() {
+    const { columns, isStriped, className, mods, style } = this.props
+    const { items } = this.state
+
+    return (
+      <Panel className={ className } mods={ mods } isStriped={ isStriped } style={ style }>
+        <PanelBody>
+          <PanelRow isWithCells>{ columns.map(this.renderColumnHeader) }</PanelRow>
```
Ollo: Should this map be in another method? Will this constantly re-render?

PR #70, opened 2018-04-09
Add popup component and status buttons

saralohr:
# Adds `Popup` component and some button modifiers.

## `Popup`
Includes modifiers for several popup positions:
- Default - up and centered
popup-default

- Overlay - on top of the triggering element, no pointer
popup-overlay

- Down - below the triggering element, centered by default
popup-down

- Right - right-aligned with triggering element (can be combined with down modifier)
popup-right

- Right Hang - hangs off the right side of an element. Useful for small buttons or icon buttons.
popup-right-hang

- Left - left-aligned with triggering element (can be combined with down modifier)
popup-left

- Left Hang - hangs off the left side of an element. Useful for small buttons or icon buttons.
popup-left-hang

# Status buttons

Two options for status buttons.

1. Button is neutral by default, and when hovered or `is-active` (as in ButtonGroup), it gets yes/no/maybe color

button-status-1

2. Button has yes/no/maybe color by default
button-status-2

# Small square button

Button intended for symbols or icons, has fixed width and height equal to height of `Button--small`.

screen shot 2018-04-09 at 9 59 43 am

### DIFF
--- a/src/css/components/Button.scss
+++ b/src/css/components/Button.scss
@@ -1,8 +1,6 @@
 @import "../config/index.scss";
 
-/**
- * @define default Button
- */
+// * @define default Button
 
 $Button-bg: $ts-white;
 $Button-bg-hover: $cu-background;
@@ -67,9 +65,8 @@ $Button-padding: 0 scaleGrid(2);
   }
 }
 
-/**
- * Modifier: Button Primary
- */
+
+// * Modifier: Button Primary
 
 $Button--primary-bg: $cu-secondary;
 $Button--primary-bg-hover: scaleColor($Button--primary-bg, -1);
@@ -107,9 +104,8 @@ $Button--primary-border-color-hover: scaleColor($cu-secondary--dark, -1);
   }
 }
 
-/**
- * Modifier: Button Negative
- */
+
+// * Modifier: Button Negative
 
 $Button--negative-bg: $ts-white;
 $Button--negative-bg-hover: $cu-negative;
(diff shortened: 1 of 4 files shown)


src/css/components/Button.scss
```diff
@@ -1,8 +1,6 @@
 @import "../config/index.scss";
 
-/**
- * @define default Button
- */
+// * @define default Button
```
saralohr: Just changing markup so docs aren't included in the compiled CSS.

src/css/components/ButtonGroup.scss
```diff
@@ -18,8 +18,11 @@ $ButtonGroup-space: -1 * $border-width-small;
 // * 1. Overlap elements and override any vertical margins.
 // * 2. Reset `border-radius` so corners line up.
 
+.ButtonGroup > .Button + .Button {
+  margin-left: $ButtonGroup-space; // 1
```
saralohr: The way this was done previously (adding negative right and left margins to all buttons in button group) was overkill. We only need it on one side of adjacent buttons to make it so the border isn't doubled up.

src/css/components/Button.scss
```diff
@@ -166,9 +161,17 @@ $Button--small-padding: 0 $su-small;
   padding: $Button--small-padding;
 }
 
-/**
- * Modifier: Button Large
- */
+// * Modifier: Button Square small
```
williamdingwall: nice, I like this for the icon buttons

saralohr: If in the future we want default and large icon buttons, we can add those too.

PR #73, opened 2018-04-09
Add loader icon

saralohr:
Adds new loader icon and modifer class `Icon--loader` to apply animation.

loader-icon-2
loader-icons


src/css/components/Icon.scss
```diff
@@ -17,3 +17,44 @@
...
+
+  circle {
+    stroke-dasharray: $Icon--loader-offset;
+    stroke-dashoffset: 0;
+    transform-origin: center;
+    animation: dash $Icon--loader-duration ease-in-out infinite
+  }
+}
+
+@keyframes rotator {
```
williamdingwall: I have one in the loading icons that does the exact same thing... might be good to consolidate these keyframe animations at some point

saralohr: Ah, yeah. That would be nice, like a new file in config or something.

williamdingwall: yeah

PR #74, opened 2018-04-10
adding feedback component and cleaning up progress

williamdingwall:


mikereinmiller:
👍 I like it!  I think it was a good idea to break the Progress and Radial progress bars apart.  :shipit:

PR #78, opened 2018-04-12
Modify component colors to fix themes

saralohr:
- Modify links color variable to match what is used in themes and Classic
- Update button colors to work with themes
- Update Panel and Popup colors to work with themes

### DIFF
--- a/src/css/components/Button.scss
+++ b/src/css/components/Button.scss
@@ -2,10 +2,16 @@
 
 // * @define default Button
 
-$Button-bg: $ts-white;
-$Button-bg-hover: $cu-background;
-$Button-color: $cu-info;
-$Button-border-color: $cu-divider;
+// Global default button colors that are themed
+$button-color:                    $cu-foreground !default;
+$button-border-color:                $cu-divider !default;
+$button-text-color:                  $cu-info !default;
+
+// Default button styles
+$Button-bg: $button-color;
+$Button-bg-hover: scaleColor($Button-bg, -1);
+$Button-color: $button-text-color;
+$Button-border-color: $button-border-color;
 $Button-border-color-hover: scaleColor($Button-border-color, -1);
 $Button-border-radius: $border-radius-medium;
 $Button-border-width: $border-width-small;
@@ -43,6 +49,7 @@ $Button-padding: 0 scaleGrid(2);
   &:focus {
     background-color: $Button-bg-hover;
     border-color: $Button-border-color-hover;
+    color: $Button-color;
     outline: none;
   }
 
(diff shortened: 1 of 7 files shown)


src/css/components/Input.scss
```diff
@@ -46,3 +47,11 @@ $form-input-bg-color: $cu-foreground--dark !default;
...
   }
 }
+
+// Hack to overrid webkit autofill styles
+input:-webkit-autofill,
+input:-webkit-autofill:hover,
+input:-webkit-autofill:focus {
+  -webkit-box-shadow: 0 0 0px 1000px $cu-foreground--dark inset;
+  -webkit-text-fill-color: $cu-info;
+}
```
saralohr: What do you think about this @williamdingwall? It overrides that grody yellow background color that Chrome adds when fields are autofilled. I think this is a user preference, so it feels weird, but I don't ever even remember setting it, and it looks pretty bad with some of our themes and with our inset shadow.

src/css/components/Button.scss
```diff
@@ -111,7 +121,7 @@ $Button--negative-bg: $ts-white;
 $Button--negative-bg-hover: $cu-negative;
 $Button--negative-color: $cu-negative;
 $Button--negative-color-hover: $ts-white;
-$Button--negative-border-color: $cu-divider;
+$Button--negative-border-color: #d6d6d6;
```
williamdingwall: why change to the hard coded value?

src/css/components/Input.scss
williamdingwall: I'm cool with re-styling it.. yeah I doubt most people, or anyone except for  a veeeery small amount actually set that.

src/css/components/Popup.scss
```diff
@@ -1,3 +1,6 @@
+$Popup-bg-color: $cu-foreground;
```
williamdingwall: I like having these settings up here, but kinda think unless they're set to `!default` it might just be an extra step.  The value of these variables is to provide people with the ability to make changes through a settings file rather than the CSS itself.  When we discuss theming let's circle back on these element variables

saralohr: Oh, right - I still see them being set on a settings file level. I can add `!default` to these.

src/css/components/Button.scss
saralohr: The negative button was looking funky on some themes, and since the other colors are fixed (we don't theme $cu-negative I don't think, or we shouldn't) I was thinking the negative button would just be fixed...  I guess I was probably viewing it out of context, though, like on a white background instead of a dark background. I can revert this.

saralohr:
@williamdingwall updated - `!default` added to Popups and negative button border reverted

PR #79, opened 2018-04-12
TS-2695: Adds data-tracking-label to Button.

seesarahcode:
- Adds ability to use a `data-tracking-label` in `Button`--ideally would like to use `data-*` instead, so feedback appreciated!

### DIFF
--- a/src/js/components/Button/Button.js
+++ b/src/js/components/Button/Button.js
@@ -51,7 +51,7 @@ class Button extends PureComponent {
   }
 
   renderButtonLink = () => {
-    const { routerLink, location, style, isDisabled, onClick } = this.props
+    const { routerLink, location, style, isDisabled, onClick, dataTrackingLabel } = this.props
 
     return (
       <TextLink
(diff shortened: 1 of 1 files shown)


saralohr:
I'm adding @mikereinmiller to this PR @seesarahcode We've been chatting about how to make this prop option more generic.

mikereinmiller:
🤔 I like the idea here and where this is going, but I have a few concerns related to having such a specific use case on these generic components.  

We looked at doing this earlier, but we weren't sure on the use case so we simplified it for now.  I think it would be beneficial to maybe pass something like `otherProps` as an object to the component.  We could then take whatever is passed into `otherProps` and spread it directly on the element like `{ ...otherProps }`.    Also using a generic name like otherProps, we could start to update the other components with a similar pattern, that way we have some similar props in the api.  Like className, mods, style, otherPros, etc.

mikereinmiller:
I'm guessing the changes made in https://github.com/teamsnap/teamsnap-ui/pull/80 take care of this PR?  If so, we could probably close this one?

saralohr:
Yep, thanks Mike.

PR #80, opened 2018-04-13
Add generic button props

saralohr:


saralohr:
cc @mikereinmiller @Ollo adding an option for generic props on the button.

PR #82, opened 2018-05-01
Adjust SummaryList Component & Styles

mikereinmiller:
I updated the SummaryList component and took a stab at adjusting some of the styles to better suite some recent designs.   If this went in the wrong direction I can remove the styles and adjust the component accordingly.  just thought I would take a stab at it.

Below is the adjustments that were made.  Some smaller spacing and adjustments to border-size if no heading is present as well as adding some style around the footer. 

### Update:

I've ensured @williamdingwall's feedback has beed added and I updated the prop behavior to provide both styles of footer text.

### DIFF
--- a/src/css/config/config.scss
+++ b/src/css/config/config.scss
@@ -125,6 +125,7 @@ $tu-bold-fontWeight:      700 !default;
 $tu-xsmall-fontSize:           scaleFont($fontSize, -2) !default;
 $tu-small-fontSize:            scaleFont($fontSize, -1) !default;
 $tu-base-fontSize:              scaleFont($fontSize, 0) !default;
+$tu-medium-fontSize:            scaleFont($fontSize, 1) !default;
 $tu-large-fontSize:             scaleFont($fontSize, 2) !default;
 $tu-xlarge-fontSize:            scaleFont($fontSize, 5) !default;
 $tu-xxlarge-fontSize:           scaleFont($fontSize, 6) !default;
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,27 +1,27 @@
 {
   "name": "@teamsnap/teamsnap-ui",
-  "version": "2.10.0",
+  "version": "2.13.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.0.0-beta.44",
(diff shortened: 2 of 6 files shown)


williamdingwall:
@mikereinmiller your changes are looking good! - there are more changes that could be made to get it closer to the comp:

screen shot 2018-05-01 at 10 56 06 am

- The heading font size drops $tu-base-fontSize
- The top border width is $border-width-medium
- The interior section dividers are $border-width-small
- The leaders between the text and amount are "dotted" instead of dashed
- All border colors here are $cu-divider
- 2-4px of spacing between the leaders and text would be nice too

Ollo:
https://github.com/teamsnap/teamsnap-ui/network/dependencies

src/css/config/config.scss
```diff
@@ -125,6 +125,7 @@ $tu-bold-fontWeight:      700 !default;
 $tu-xsmall-fontSize:           scaleFont($fontSize, -2) !default;
 $tu-small-fontSize:            scaleFont($fontSize, -1) !default;
 $tu-base-fontSize:              scaleFont($fontSize, 0) !default;
+$tu-medium-fontSize:            scaleFont($fontSize, 1) !default;
```
saralohr: One thought to skirt the issue - could the footer text work as the `base` size, and let the bold font weight distinguish it?

If not...

This one is tricky...trying to make a numerical size scale arbitrarily align with a sm, md, lg scale. We equated base with medium on our text size utilities. Unfortunately something like `xmedium` or `Xm`, while consistent with small and large, doesn't make a whole lot of sense.

screen shot 2018-05-07 at 2 58 45 pm

We could keep what you have, and update the text utilities to reflect that change. I do think there is value in having a class labeled as the base size if you want to revert back it.

Sorry, I am thinking this out while writing. If using one of the existing sizes doesn't work I think it's okay to keep what you have for now, and Bill and I probably need to rethink the text scale.

mikereinmiller: I'm totally flexible in how this is done.  If it's easier for now, I can use the base and set it apart with a bold font-weight?  The we can talk about how maybe these font-sizes should operate in the long run?

saralohr: > If it's easier for now, I can use the base and set it apart with a bold font-weight

Yeah, let's just do that for now. Thanks!

PR #83, opened 2018-05-09
Add new color ways and add huge size

l3aconator:


l3aconator:
@saralohr Thank you for the suggestions, updated it!

PR #86, opened 2018-05-12
Add textarea to is-notValid style class

amberwebb:
When need to have the red border when validating textarea.

image


saralohr:
I've never understood exactly what package-lock.json does, but I _think_ it's updated there too when you run the `npm version` command to update the version rather than doing it manually.

PR #90, opened 2018-05-31
Add initial changelog and update contributing directions

saralohr:


CONTRIBUTING.md
```diff
@@ -11,7 +11,9 @@ Once your user is created and added to the org this guide details the steps to p
 * rebase and get  your branch ahead of master
 * then from your working branch
 * `npm login --scope=@teamsnap`
-* `npm version major/minor/patch -m “commit message”`
+* `npm version major/minor/patch`
```
Ollo: Its important to include the `-m  "commit message"` part otherwise the version update is not automatically applied as part of the commit history and you have to manually update the version numbers.

saralohr: Ah, okay. I was wondering about that. I'll change that back.

PR #91, opened 2018-06-01
Add wrapper around table rows with height prop

amberwebb:
We need the Table component body to be scrollable, while the table header is stationary. This PR adds a wrapper around the table row function where height and overflow are passed as inline styles. The value for height is `tableBodyHeight` prop. The default for `tableBodyHeight` is `auto`.

@saralohr Is this what you had in mind?

### DIFF
--- a/src/js/components/Table/Table.js
+++ b/src/js/components/Table/Table.js
@@ -151,9 +151,7 @@ class Table extends PureComponent {
   }
 
   render() {
-    const { isStriped, className, mods, style, otherProps } = this.props
-
-    const tableBodyHeight = this.props.tableBodyHeight || 'auto'
+    const { isStriped, className, mods, style, otherProps, tableBodyHeight } = this.props
 
     return (
       <Panel className={ className } mods={ mods } isStriped={ isStriped } style={ style } { ...otherProps }>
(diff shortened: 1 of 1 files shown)


src/js/components/Table/Table.js
```diff
@@ -151,13 +151,15 @@ class Table extends PureComponent {
...
   render() {
-    const { isStriped, className, mods, style, otherProps } = this.props
+    const { isStriped, className, mods, style, otherProps, tableBodyHeight } = this.props
 
     return (
       <Panel className={ className } mods={ mods } isStriped={ isStriped } style={ style } { ...otherProps }>
         <PanelBody>
           <PanelRow isWithCells>{ this.renderTableColumns() }</PanelRow>
-          { this.renderTableRows() }
+          <div style={{height: tableBodyHeight, overflow: 'scroll' }}>
```
saralohr: We may want to only add this wrapper and styles if we want the body to scroll. Sometimes the `overflow` property can act in unexpected ways.

mikereinmiller: yeah, might have to add this as some kind of prop.  probably default to not setting the overflow height?  🤔

amberwebb:
Closing for https://github.com/teamsnap/teamsnap-ui/pull/92

PR #92, opened 2018-06-04
Update items in local state when row props change and scroll table body

amberwebb:
Table component `items` state is only set when the component mounts, not when props update.

This PR adds a conditional check in `componentDidUpdate` method and updates `items` state accordingly. Also adds a div around table body for scrolling.

### DIFF
--- a/src/js/components/Table/Table.js
+++ b/src/js/components/Table/Table.js
@@ -47,6 +47,14 @@ class Table extends PureComponent {
     this.setState({ ...tableState })
   }
 
+  componentDidUpdate(prevProps, prevState) {
+    const { rows } = this.props
+
+    if (prevProps.rows.length !== rows.length) {
+      this.setState({items: setUniqueId(rows)})
+    }
+  }
+
   sortItems = (newSort, newItems) => {
     const { columns } = this.props
 
--- a/src/js/components/Table/Table.js
+++ b/src/js/components/Table/Table.js
@@ -47,11 +47,11 @@ class Table extends PureComponent {
     this.setState({ ...tableState })
   }
 
-  componentDidUpdate(prevProps, prevState) {
+  componentDidUpdate(prevProps) {
     const { rows } = this.props
 
     if (prevProps.rows.length !== rows.length) {
-      this.setState({items: setUniqueId(rows)})
+      this.setState({ items: setUniqueId(rows) })
     }
   }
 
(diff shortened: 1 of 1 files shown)


src/js/components/Table/Table.js
```diff
@@ -157,7 +165,14 @@ class Table extends PureComponent {
       <Panel className={ className } mods={ mods } isStriped={ isStriped } style={ style } { ...otherProps }>
         <PanelBody>
           <PanelRow isWithCells>{ this.renderTableColumns() }</PanelRow>
-          { this.renderTableRows() }
+          {otherProps.scrollTableBody &&
```
mikereinmiller: This is probably not the way to go about this.   The otherProps are used to spread out on the parent component as you can see above `{ ...otherProps }`.  In this case, you'll probably want to add a new prop for this.

mikereinmiller: maybe if a set table height is passed, enable the scrollable table?   As long as if the table is less than the height it wouldn't be messed up?

amberwebb:
@mikereinmiller Another look please?

src/js/components/Table/Table.js
```diff
@@ -185,7 +200,8 @@ Table.propTypes = {
   className: PropTypes.string,
   mods: PropTypes.string,
   style: PropTypes.object,
-  otherProps: PropTypes.object
+  otherProps: PropTypes.object,
+  maxTableHeight: PropTypes.number
```
mikereinmiller: is this actually a number?  Would I be able to do, `400px` or `40em` or something?

amberwebb: Yes this needs to be a string for just that purpose.

src/js/components/Table/Table.js
```diff
@@ -47,6 +47,14 @@ class Table extends PureComponent {
     this.setState({ ...tableState })
   }
 
+  componentDidUpdate(prevProps) {
+    const { rows } = this.props
+
+    if (prevProps.rows.length !== rows.length) {
```
mikereinmiller: Could there be an issue here with the table not updating correctly?  Just wonder if the filtered results might end up having the same length, what would happen?  Am I overthinking here?  I'm not sure if we would want it to "always" update though.  @Ollo you have any thoughts on this?

mikereinmiller: I'm also curious if we want a componentWillReceiveProps instead?

Ollo: Hmm.. This feels a bit like a 🔨to me as a prop or state change should automatically trigger a re-render. Have we tested converting this from a `PureComponent` to see if we are getting bitten by the shallow compare?

amberwebb: Not sure what the hammer means... 

The component is only updating state when the component mounts, which is only one time. It should happen when it updates as well as long as something has changed.

Good call about testing out converting from PureComponent. I will try that. Perhaps that's also the reason why sorting is not working properly.

I need the table to update when there are changes, otherwise how can I display filtered items?

I really hope that as this codebase grows we can introduce tests.

amberwebb: image

amberwebb: Here, the table is only passing 1 row prop, which is correct because I am filtering on `Non Player`. `state.items`, which are the row items, are still holding the original 13 objects that it initialized with.

PR #93, opened 2018-06-05
Table Component Updates

mikereinmiller:
Updates the table component to receive new props.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - `Fixed` for any bug fixes.
 - `Security`
 
+## [2.10.0] - 2018-06-06
+https://github.com/teamsnap/teamsnap-ui/pull/93
+### Fixed
+- Fix Table component state to update when props change.
+### Added
+- Add `maxTableHeight` prop to Table component.
+
+
 ## [2.9.0] - 2018-05-31
 https://github.com/teamsnap/teamsnap-ui/pull/88
 ### Changed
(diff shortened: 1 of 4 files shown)


amberwebb:
@mikereinmiller What happens in `componentWillRecieveProps` if you do not compare this.props and nextProps? Won't this component update even if there are no changes without the comparison?

I used `componentDidUpdate` in my PR because of the recommendation here https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops

amberwebb:
When is it appropriate to use `componentDidUpdate` vs `componentWillReceiveProps`?

Even though the new lifecycle methods would cause issues in some of our apps on older versions of React, we need to think about moving away from `Will` methods since React is deprecating them. The `Did` methods are available in older versions of React. This is a conversation we should have as a Guild. It will be time consuming to go through each component and update deprecated methods, especially since we have no tests (hoping we will by that point).

But, I linked this into my Hydra branch and filtering is working as expected. Sorting is still buggy. Perhaps it's my implementation of the Table component.

mikereinmiller:
It's not really a matter of 'not updating'.  We will definitely be updating.  But these are soon going to be marked as deprecated and then in another future release they will be removed.  We aren't even to the version of getting the deprecated warnings yet.  So we have time to migrate the appropriate applications over as well as all of teamsnap-ui.  LFE is still using both the toolkit and UI components. We will be updating LFE hopefully soon or just moving it to hydra altogether.  So I think for now in the short term, it is a little early to change.  We can change all of UI at the same time, when we migrate some of the apps that are depending on these early releases still.

mikereinmiller:
@amberwebb I think those changes for max height look good!  I say have @Ollo double check over the recent changes and if he gives a 👍 we should be good.

PR #95, opened 2018-06-14
Use // to work for both http and https

saralohr:
I just noticed that the fonts weren't working in UI-Patterns with Classic CSS turned off. I only tested this locally on http, but it needs to work on https too. Using `//dugout` so it works for both, unless there is some better way to do this?


Ollo:
We can give this a try but not sure the auto switch to https will happen with this method cross domain. We may need to use JS to determine protocol or just always use https. 

**ref:** https://stackoverflow.com/a/2725534/1210367

saralohr:
Okay, I'm gonna give this a try and we'll see!

PR #96, opened 2018-07-20
adding base tab CSS

williamdingwall:
@saralohr I'm curious to here your opinion on the data attributes to control  functionality of the tabs... this is how it works within classic. Thinking it might be good to some of the functionality for classic for development in places like registration, maybe account, potentially other places.


src/css/components/index.scss
```diff
@@ -1,19 +1,20 @@
...
 @import "StepNav";
-@import "Panel";
-@import "Loader";
-@import "Divider";
-@import "Progress";
 @import "SummaryList";
+@import "Tabs";
+@import "Toggle";
 @import "Tooltip";
-@import "Feedback";
```
saralohr: Thanks for alphabetizing these.

src/css/components/Tabs.scss
```diff
@@ -0,0 +1,67 @@
...
+    display: block;
+    width: 100%;
+    height: $border-width-large;
+    position: absolute;
+    left: 0;
+    bottom: 0 - $border-width-large;
+    background: transparent;
+  }
+
+  &.js-active,
```
saralohr: This would be `is-active` according to SUIT conventions.

williamdingwall: Oh right, yeah I'll update that - here's the ui-patterns PR https://github.com/teamsnap/ui-patterns/pull/56

src/css/components/index.scss
williamdingwall: no problem!

PR #108, opened 2019-03-15
Teamsnap ui component updates

williamdingwall:
a few component and utility updates for the DMV demo

UI-patterns PR https://github.com/teamsnap/ui-patterns/pull/71


src/css/utils/size.scss
```diff
@@ -12,7 +12,7 @@ $responsive: true;
 @each $col in $cols {
   @for $i from 1 through ($col - 1) {
     .u-size#{$i}of#{$col} {
-      flex-basis: auto !important;
```
williamdingwall: @saralohr wanted to get your thoughts on this change. Was running into issues with u-sizeFit not working in on things like Panel-cell that had the flex: shorthand definition, so just overriding flex-basis wasn't enough.

saralohr: I'd be pretty hesitant to make this change, since the flex shorthand for growing and shrinking can act in unexpected ways, and it would be a divergence from the SUIT CSS system that this is pulled from. Can the Panel-cell flex property be changed instead? Could you share the specific use case of what you are trying to achieve?

williamdingwall: I am too. The issue with Panel-cell is that it's set to automatically fit the within the row with `flex: 1 1 0%;` (which I wouldn't want to change) and widths can't be defined as anything smaller than flex defines them as. It seems like the content around it can be set wider and it'll in turn shrink the cells around it to accommodate, but I'm not able to set the small columns and let the larger one grow to fit the space left. It would be nice to be able to define the smaller column since it's usually more defined content in terms of size.

saralohr: I think there may be a few existing utility classes that you could add to `Panel-cell` to get around this. Size utilities should work to override the flex-basis, or even `u-flexInitial` which sets the flex shorthand as `0 1 auto` so it won't grow to fill the space. We could also add a class `u-flexGrow0`. If not let's pair and see if we can find a solution.

williamdingwall: Good call on the u-flexGrow0 utility! I'll do that

src/css/components/Card.scss
```diff
@@ -0,0 +1,51 @@
+@import "../config/index.scss";
+
+//------------------------------------------------------------//
+
+// * Card Component
+// * 1. Define Default Card
```
saralohr: `Card` seems very similar in style to `Panel`. Do we need both?

src/css/components/Popup.scss
```diff
@@ -188,6 +188,43 @@ $Popup-arrow-borderRadius: 3px;
   }
 }
 
+// * Hover modifier
+//
```
saralohr: Nice!

src/css/utils/text.scss
```diff
@@ -70,3 +75,31 @@
 .u-fontSizeXxl {
   font-size: $tu-xxlarge-fontSize !important;
 }
+
+/**
+ * Font weight utilities
+ */
```
saralohr: Yay! Been wanting these for a while.

src/css/components/Card.scss
williamdingwall: True, I'm just wondering if we'd like to make it it's own to keep formatting consistent across the app. If it's just panel, design details could vary a fair amount over time. These are the three I'm looking at right now. The third would be a modifier...

cards

Initially I was thinking of Material's card design, but they don't have a panel. https://material.io/design/components/cards.html#anatomy

saralohr: Hrm... if there were some really specific font / layout styles just for cards I think it could make sense, but the styles added here don't quite  justify it IMO, since `.Card-content, -contentRow, and contentHeader` are all just adding minimal spacing. `Card-image` is the most compelling case for making it a component.

If you do make it it's own component, I'd suggest reducing the child classes that just add spacing to make it more versatile and easy to use.

src/css/components/Card.scss
```diff
@@ -0,0 +1,51 @@
...
+  padding: $su-medium;
+}
+
+.Card-contentHeader {
+  padding-bottom: $su-small;
+}
+
+.Card-contentRow {
+  padding-bottom: $su-medium;
+  list-style: none;
```
saralohr: Will row always be a list? Might be a good opportunity to add a `u-listUnstyled` utility class.

src/css/components/Card.scss
williamdingwall: Fair points. I think if scaling back the card element to focus on the image, I might lean towards just doing an Image component or utilities. List too. I'll do that for now... It could be too that this level of component should be in React and not the CSS. My concern in general with scaling back to smaller components is that we're giving devs puzzle pieces that are less intuitive to put together.

saralohr: If you do stick with Panel, FWIW, I think `Panel-image` would be a useful child of Panel, though we may want variations on the max-height and how the image is positioned within the container (for example, center-aligned vs top-aligned) OR those attributes could be great candidates for adding on the React component based on context.

williamdingwall: Yeah it's hard to say. A Panel image wouldn't really work on large panels and I feel like a child element should work at all sizes. Kinda back and forth on it all too.

src/css/components/Card.scss
williamdingwall: No not necessarily - good call, I'll do that

williamdingwall: I just got rid of the list for now, will probably come back and do a list component or utilities. We have a few styles that would be nice to add here. Like comma separated.

williamdingwall:
Thanks @saralohr  went back and got rid of card

PR #112, opened 2019-05-01
Popup focus state & Table Knobs

williamdingwall:
- adding a focus state and focus within state for .Popup--hover modifier
- adding table knobs to storybook
- updating change log to better reflect recent work

popup-tabbing

table-knobs

### DIFF
--- a/src/js/components/Table/Table.stories.js
+++ b/src/js/components/Table/Table.stories.js
@@ -1,20 +1,33 @@
 import React from 'react';
 import { storiesOf } from '@storybook/react';
-// import { withKnobs, text, boolean, number, selectV2 } from '@storybook/addon-knobs/react';
+import { boolean, select } from "@storybook/addon-knobs/react";
 import Table from './Table';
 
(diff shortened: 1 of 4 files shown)


src/js/components/Table/Table.stories.js
```diff
@@ -1,20 +1,33 @@
 import React from 'react';
 import { storiesOf } from '@storybook/react';
-// import { withKnobs, text, boolean, number, selectV2 } from '@storybook/addon-knobs/react';
+import { boolean, select } from "@storybook/addon-knobs/react";
 import Table from './Table';
 
  const stories = storiesOf('Table', module);
 
+ const columns = [
```
williamdingwall: @dmccraw just wanted to run this update by you to see if there's a better way to do this, or if it looks good to you?

dmccraw: this looks good to me. Thank you for getting started! I like having const's outside the render so it doesn't have to reallocate them on every render

PR #123, opened 2019-06-04
added select stories.js

TimWatervoort:
This adds a storybook element for the Select component.

It has two stories:
 * Default, which only has a standard select box with three options
 * Alternates, which has a large select box, and a select box with one disabled option

### DIFF
--- a/src/js/components/Select/Select.stories.js
+++ b/src/js/components/Select/Select.stories.js
@@ -1,0 +1,57 @@
+import React from "react";
+import { storiesOf } from "@storybook/react";
+import Select from "./Select";
+
+const stories = storiesOf("Select", module);
+
+const optionsList = [
+  {
+    label: 'Option 1',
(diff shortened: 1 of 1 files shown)


src/js/components/Select/Select.stories.js
```diff
@@ -0,0 +1,57 @@
...
+  },
+  {
+    label: 'Option 2',
+    value: 'option 2'
+  },
+  {
+    label: 'Option 3',
+    value: 'option 3'
+  }
+]
```
dmccraw: would it be nice to add sub menus in the select?

TimWatervoort: The current React component doesn't support that, do you think that is something we should look at adding in?

dmccraw: if it's not supported then don't worry about it.

PR #124, opened 2019-06-05
Panel story

williamdingwall:
adding a panel story and panelImage proptype

### DIFF
--- a/src/js/components/Panel/Panel.js
+++ b/src/js/components/Panel/Panel.js
@@ -18,7 +18,7 @@ import { getClassName } from '../../utils/helpers'
 
 class Panel extends PureComponent {
   render () {
-    const { children, className, mods, isStriped, maxSize, panelImage, style, otherProps } = this.props
+    const { children, className, mods, isStriped, maxSize, style, otherProps } = this.props
 
     const panelClasses = getClassName(
       className,
@@ -29,11 +29,6 @@ class Panel extends PureComponent {
 
     return (
       <div className={ panelClasses } style={ style } { ...otherProps }>
-        {panelImage && 
-          <div className='Panel-image'>
-            <img src={panelImage.Source ? panelImage.Source : panelImage.Placeholder } />
-          </div>
-        }
         { children }
       </div>
     )
@@ -44,7 +39,6 @@ Panel.propTypes = {
   children: PropTypes.node.isRequired,
   isStriped: PropTypes.bool,
   maxSize: PropTypes.string,
-  panelImage: PropTypes.object,
   className: PropTypes.string,
   mods: PropTypes.string,
   style: PropTypes.object,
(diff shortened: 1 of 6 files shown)


williamdingwall:
@dmccraw @amy-r I'm curious on your thoughts with this approach. Wanted to keep 'panelImage' down to a single prop so set it up to take an object containing a an image src and placeholder source.

src/js/components/Panel/Panel.stories.js
```diff
@@ -0,0 +1,102 @@
...
+            on to star in comedy films. He also co-directed Quick Change (1990). Murray garnered additional critical acclaim later 
+            in his career, starring in Lost in Translation (2003), which earned him a Golden Globe and a BAFTA Award for Best Actor, 
+            as well as an Academy Award nomination for Best Actor. He also received Golden Globe nominations for his roles in 
+            Ghostbusters, Rushmore (1998), Hyde Park on Hudson (2012), St. Vincent (2014), and the HBO miniseries Olive Kitteridge 
+            (2014), for which he later won his second Primetime Emmy Award for Outstanding Supporting Actor in a Limited Series or a Movie.
+            </p>
+          </PanelCell>
+          <PanelCell mods='Grid-cell'>
+              <img src='https://www.fillmurray.com/640/360' />
+          </PanelCell>
```
dmccraw: is there a reason you aren't using panelImage here?

src/js/components/Panel/Panel.js
```diff
@@ -39,6 +44,7 @@ Panel.propTypes = {
   children: PropTypes.node.isRequired,
   isStriped: PropTypes.bool,
   maxSize: PropTypes.string,
+  panelImage: PropTypes.object,
```
dmccraw: I wonder if it's worth calling out what keys the panelImage object has.
`
PropTypes.shape({Source, PlaceHolder})
`
?

src/js/components/Panel/Panel.stories.js
williamdingwall: Yeah an image in a panel is different from a "panel image". I updated it to headerImage in the next commit to make it a little clearer.

Screen Shot 2019-06-05 at 1 46 11 PM
Screen Shot 2019-06-05 at 1 46 18 PM

On it's just content, the other the image become more a part of the panel

src/js/components/Panel/Panel.js
williamdingwall: Oh yeah I like that, that's much clearer

src/js/components/Panel/Panel.stories.js
```diff
@@ -0,0 +1,102 @@
...
+      <PanelFooter>
+        <span className='Button Button--negative'>Cancel Button Placeholder</span>
+        <span className='Button Button--primary u-spaceLeftSm'>Confirm Button Placeholder</span>
+      </PanelFooter>
+  </Panel>
+))
+
+stories.add('Panel Image', (() =>
+<Panel 
+  panelImage={ select('Select Image', imageOptions, noImage) }
```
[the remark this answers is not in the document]
williamdingwall: updated to use shape so there's more of an indication of what's expected

PR #127, opened 2019-06-06
Added Feedback component/story

TimWatervoort:
Added a Feedback component to the ui.
Added a story that describes the various props that the Feedback component can take.


dmccraw:
@jen-prince @TimWatervoort Sounds like there needs to be a discussion around the feedback component.

dmccraw:
@TimWatervoort @jen-prince Was there every a discussion around the feedback component?

PR #128, opened 2019-06-24
Add in JSX rendering story for table

bradcypert:
I came in to modify the table so that it could render JSX elements instead of just strings and was pleasantly surprised to see that it already worked that way! I did end up formatting the table file as well as adding a new storybook data set that showcases using JSX instead of strings.


dmccraw:
@bradcypert It looks like your editor is not picking up the linting rules associated with this repo. You might want to run `eslint --fix` so that we can only see the changes.

bradcypert:
Ran eslint fix and it fixed quite a few files, including ones that I hadnt touched. More files are in there now, but they should all be formatted as expected.

PR #130, opened 2019-07-11
Migrate To Typescript

bradcypert:
This PR takes the steps necessary to get teamsnap-ui working politely with typescript. It made sense to me to make sure that, if most every team would use this library, then it should try it's best to support those teams -- particularly via TypeScript definitions for teams using TypeScript.

Additionally, this is configured so that our components will work with storybook.


williamdingwall:
@bradcypert how will this effect repos like League-Frontend and invoicing-frontend that use the current js components and not typescript?

bradcypert:
@williamdingwall great question! TypeScript is only the pre-compiled language. Browsers cant currently run TypeScript so what the build process looks like is this:
1. We take all typescript files and run them through the typescript compiler. This step produces JavaScript and JSX.
2. The emitted JavaScript (to my best knowledge) matches the patterns in place for the existing JavaScript. It still works with the Storybook example too! :)

I'm glad you mentioned both of those frontends. I'll want to pull those and manually replace teamsnap-ui with this branch's build output just to test and make sure everything is still A-Okay, but it definitely should be. Simply by supporting TypeScript at the teamsnap UI layer, we get better support for using Teamsnap-UI in TypeScript projects, but you can still use it in JavaScript exactly like before.

bradcypert:
@williamdingwall I tested League frontend and noticed that it was having trouble with loaders and some of the files. I've changed the export mode to commonjs which is what league frontend (and the other frontends from the looks of it) are expecting. I was able to build the typescript version of teamsnap-ui locally and wire it up to league frontend with much success!

williamdingwall:
Thanks for checking on that @bradcypert ! and thanks for the explanation, sounds good to me. What's going on with the loader? Did that get fixed with the change to commonjs?

bradcypert:
Yup! All I had to do was export it as CommonJS instead of ES6.

mikereinmiller:
This is some nice work getting all of this migrated to Typescript.  No small task, but I'm really hesitant to start migrating our stuff in general to typescript.  Personally I'm not fully sold on it yet, especially when we start comparing it against something like Flow (since we are using React moving forward).  Also for something like this we are increasing our overall bundle size with typescript as opposed to raw JS, so we'll need to be aware of that as well. We can talk more about this in the FE guild meeting.  But either way, nice work.

dmccraw:
We have any metrics on the size difference between the pure JS version and the Typescript version of the library?

bradcypert:
I dont know if we're actually going to increase the bundle size much, if at all simply because the typescript code gets ejected by the compiler and we only send a JS bundle to the browser. I didn't profile the before and after, but that shouldnt be too difficult to do.

bradcypert:
@mikereinmiller Also, one of the big benefits to this being migrated to TypeScript is that our JavaScript teams can still use it as well, but we can better support our TypeScript teams by using TypeScript at the library-level (like TSUI). This ensures that teams that want TypeDefs can have them, but teams that don't want to worry about TypeDefs can just use the plain ol' JS.

bradcypert:
Wanted to share the file-size adjustments as well.

## Before TypeScript
Screen Shot 2019-08-06 at 5 28 58 PM

## After TypeScript
Screen Shot 2019-08-06 at 5 26 47 PM

It's worth mentioning that __with__ the typescript definitions, the file size will be bigger, but those aren't shipped with the application and are consumed by the compiler. Interestingly enough, as far as what does get shipped to the browser, we've shaved off 4k by switching to TypeScript.

PR #154, opened 2019-09-05
Variables utilities updates

williamdingwall:
1. Refactoring theme variables to separate the config from the compiled CSS, so the config can be imported directly in other apps. 

example: https://github.com/teamsnap/organization-frontend/blob/master/src/styles/main.scss#L1

2. Adding a few utilities for a few different projects going on right now

### DIFF
--- a/src/css/utils/border.scss
+++ b/src/css/utils/border.scss
@@ -81,3 +81,20 @@ $border-radius: (
     }
   }
 }
+
+// Border Color
+
+$border-radius: (
+  'Default': $cu-divider,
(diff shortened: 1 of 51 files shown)


src/css/utils/border.scss
```diff
@@ -81,3 +81,20 @@ $border-radius: (
     }
   }
 }
+
+// Border Color
+
+$border-radius: (
```
saralohr: Should this be `$border-color`?

williamdingwall: nice catch, thanks!  updated

PR #162, opened 2019-09-09
Add in components for PopUp buttons

bradcypert:
https://teamsnap.atlassian.net/secure/RapidBoard.jspa?rapidView=242&modal=detail&selectedIssue=TNL-1880

Organization is using an existing popup that is merged into TeamSnap UI in this PR called PopupActions. This also includes PopupConfirm, which presents the user with a yes or no confirm option, like in https://teamsnap-ui-patterns.netlify.com/patterns/components/popup.html#overlay

### DIFF
--- a/src/js/components/Popup/PopupAction.tsx
+++ b/src/js/components/Popup/PopupAction.tsx
@@ -1,0 +1,89 @@
+import * as React from "react";
+
+interface State {
+  isPopupOpen: boolean;
+}
+
+interface Action {
+  text: string;
+  callback: () => void;
+}
+
+interface Props {
+  text: string | React.ReactElement;
+  actions: Array<Action>;
+  popupStyle?: React.CSSProperties;
+  direction?: "down" | "right" | "left" | "rightHang" | "leftHang" | "overlay";
+}
+
+export default class PopUpAction extends React.Component<Props, State> {
+  constructor(props: Props) {
+    super(props);
+    this.state = {
+      isPopupOpen: false
+    };
+  }
+
+  componentDidMount() {
+    document.body.addEventListener("click", this.handleBodyClick.bind(this));
+  }
+
+  componentWillUnmount() {
+    document.body.removeEventListener("click", this.handleBodyClick.bind(this));
+  }
+
+  handleBodyClick() {
+    this.setState({
+      isPopupOpen: false
+    });
+  }
+
+  togglePopup() {
+    let isPopupOpen = this.state.isPopupOpen;
+    this.setState({
+      isPopupOpen: !isPopupOpen
+    });
+  }
+
+  render() {
+    return (
+      <div className="Popup Popup--hover">
+        <button
+          className="Button Button--small"
+          onClick={this.togglePopup.bind(this)}
+        >
+          {this.props.text}
+        </button>
+        <div
+          className={
+            "Popup-container " +
+            (this.props.direction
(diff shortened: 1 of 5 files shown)


src/js/components/Popup/PopupConfirm.tsx
```diff
@@ -0,0 +1,85 @@
...
+      <div className="Popup">
+        <button
+          className="Button Button--small"
+          onClick={this.togglePopup.bind(this)}
+        >
+          {this.props.buttonText}
+        </button>
+        <div
+          className={
+            "Popup-container " +
```
williamdingwall: I like this approach for generic popup. It might make sense to drop the confirm part of the naming, since it could be used for any number of popups, like player cards in the future. Confirms would always be overlays. And again can will this able to take multiple arguments (Up, Left), (Down, Right), etc?

williamdingwall: Actually with this one since it's just a generic cancel + accept with text, maybe just go with the overlay without options. A more generic popup that accepted any content and options for style and sizing would be cool though

src/js/components/Popup/PopupAction.tsx
```diff
@@ -0,0 +1,89 @@
...
+        <button
+          className="Button Button--small"
+          onClick={this.togglePopup.bind(this)}
+        >
+          {this.props.text}
+        </button>
+        <div
+          className={
+            "Popup-container " +
+            (this.props.direction
```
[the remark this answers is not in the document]
bradcypert: I made some modifications so that we can have multiple options now.

src/js/components/Popup/PopupConfirm.tsx
bradcypert: Dropped the options and have it set to an overlay. We may want to revisit this at some point for a generic (whatever you want goes inside) pop over.

amy-r:
Are the user's tab features automatically moved to the popup and the options? How does it work with screen readers?

bradcypert:
> Are the user's tab features automatically moved to the popup and the options? How does it work with screen readers?

That's a great question. For the Confirm one, you can tab into the actions, which seems good. The PopupActions one however, doesn't do that currently. I'll need to modify the markup to handle that.

src/js/components/Popup/PopupAction.tsx
```diff
@@ -69,13 +69,20 @@ export default class PopUpAction extends React.Component<Props, State> {
...
                   <div key={action.text}>
-                    <p
+                    <button
+                      tabIndex={0}
                       className="u-padEndsSm u-padSidesMd"
-                      style={{ cursor: "pointer" }}
+                      style={{
+                        cursor: "pointer",
+                        appearance: "none",
+                        border: "none",
```
williamdingwall: @bradcypert looks like you'll need to set background to none too.

Screen Shot 2019-09-10 at 10 50 18 AM

src/js/components/Popup/PopupAction.tsx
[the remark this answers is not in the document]
williamdingwall: cool, thanks! could you add that to storybook too by chance?

src/js/components/Popup/PopupAction.tsx
bradcypert: Updtaed. What browser were you testing in?

PR #164, opened 2019-09-10
Modify popup actions to allow confirmation requirements, allow multiple directions

bradcypert:


src/js/components/Popup/PopupAction.tsx
```diff
@@ -1,26 +1,39 @@
 import * as React from "react";
+import { PopupConfirm } from ".";
```
williamdingwall: looks like you could cut this

bradcypert: good call

PR #179, opened 2019-10-03
cleanup on disabled selects, inputs, and variables

williamdingwall:
part of this ticket https://teamsnap.atlassian.net/browse/TNL-1673

cleaning up the disabled inputs and selects to make values readable. Previously the value of an input was the same color as the placeholder text.

---

Inputs:

no text entered:
Screen Shot 2019-10-03 at 2 47 51 PM

text entered:
Screen Shot 2019-10-03 at 2 48 18 PM

SelectBox:
Screen Shot 2019-10-03 at 2 48 32 PM

---

Updated to match some classic changes:
https://github.com/teamsnap/classic/pull/8860

Screen Shot 2019-10-03 at 2 07 18 PM

### DIFF
--- a/src/css/components/SelectBox.scss
+++ b/src/css/components/SelectBox.scss
@@ -6,14 +6,22 @@
 // * 1. Select Box Settings
 // * 2. Select Box
 
+// TODO: update the dropdown icon to SVG
+
 //------------------------------------------------------------//
 
 // * 1. Select Box Settings
 
-$SelectBox-bg-color: $cu-foreground--dark !default;
-$SelectBox-text-color: $cu-info !default;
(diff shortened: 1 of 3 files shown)


src/css/components/SelectBox.scss
```diff
@@ -68,11 +81,19 @@ $SelectBox-indicator-width: scaleGrid(4);
     outline: none;
   }
 
+  &:invalid {
+    color: red !important;
```
saralohr: Use `negative` color variable?

williamdingwall: oops... was testing something to get that to show, but wasn't getting it to work.. cutting

williamdingwall:
thanks!

PR #185, opened 2019-10-16
Remove deprecated lifecycle method for prep for react 17

bradcypert:
Trying to get TeamSnap UI ready for React 17+.

I think this is the only component with deprecated lifecycle methods. 

So this PR removes those lifecycle methods in favor of the alternatives proposed in https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#updating-state-based-on-props

### DIFF
--- a/src/js/components/Table/Table.tsx
+++ b/src/js/components/Table/Table.tsx
@@ -83,40 +83,35 @@ class Table extends React.PureComponent<any, State> {
 
   constructor(props) {
     super(props);
-    const { rows, defaultSort } = this.props;
+    const { defaultSort } = this.props;
 
     // Establish initial sortDirection by checking for '-' value
     const sortDirection = defaultSort.charAt(0) === "-" ? true : false;
     const sortName = sortDirection ? defaultSort.substr(1) : defaultSort;
 
-    const state = this.getTableState(rows, sortName, sortDirection);
+    const state = Table.getTableState(props, sortName, sortDirection);
     this.state = { ...state };
   }
 
-  componentWillReceiveProps(nextProps) {
-    const { sortByColumn, sortByReverse } = this.state;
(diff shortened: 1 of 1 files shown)


mikereinmiller:
Great catch @bradcypert this was on my TODO list, glad to see you were able to knock it out quickly!  👍

src/js/components/Table/Table.tsx
```diff
@@ -75,40 +81,37 @@ class Table extends React.PureComponent<any, any> {
...
       : { items };
 
-    this.setState({ ...tableState });
-  };
+    return tableState;
+  }
 
-  sortItems = (newItems, sortByColumn, sortByReverse) => {
-    const { columns } = this.props;
+  static sortItems = (props, newItems, sortByColumn, sortByReverse) => {
```
amy-r: I'm curious about which methods are private. What's the thinking behind private getTableState and not sortItems?

bradcypert: A great question, I made new methods private and left the others as is (for now). Unfortunately, we have a public library using semantic versioning, so it'd be unwise to start making public methods private without a major version bump (people might be using those methods!)

Ultimately, the rules for API design that I try to follow are as follows:
1. New methods should be private unless they otherwise need a public modifier.
2. You can always make private methods public later (that wont break anything)
3. Making public methods private requires significant overhead (or a PSA that things may break).

That being said, since we're moving sortItems to a static modifier, I guess that technically is a breaking change, so it might make sense to go ahead and make it private as well. The only downside is that if someone is using the public sort items, it's not hard to simply change the callsite to a static call, but if its private, they can no longer use it (they'll have to write their own).

dmccraw: Great notes @bradcypert I wonder if we should document some of these as a wiki page or in the readme.

PR #263, opened 2020-02-18
Add in opinionated pagination table

bradcypert:
This PR adds in a table that has wired up pagination controls and works with asynchronous data. This seems to be a common theme with several of our frontend applications, and the current table's sorting doesn't play nicely with paginated data, so this is an attempt to help remedy that situation by providing an easy to use solution.

### DIFF
--- a/src/js/components/Table/Paginated/PaginatedTable.stories.tsx
+++ b/src/js/components/Table/Paginated/PaginatedTable.stories.tsx
@@ -1,0 +1,69 @@
+import * as React from "react";
+import { storiesOf } from "@storybook/react";
+import PaginatedTable from "./PaginatedTable";
+
+const stories = storiesOf("PaginatedTable", module);
+const columns = [
+  { name: "name", label: "Member Name", isSortable: true },
+  { name: "gender", label: "Gender", isSortable: true },
+  { name: "age", label: "Age", isSortable: true },
+  { name: "programs", label: "Active Programs", isSortable: true }
+];
+
+const data = Promise.resolve<any[]>([
+  {
+    name: "Brad",
+    gender: "m",
+    age: 12,
+    position: "Goalie",
+    activePrograms: [
+      { name: "2019 ACC Academy", subtitle: "(Junior Academy Tryout)" },
+      { name: "2019 ACC Academy", subtitle: "(Junior Academy Tryout)" },
+      { name: "2019 ACC Academy", subtitle: "(Junior Academy Tryout)" },
+      { name: "2019 ACC Academy", subtitle: "(Junior Academy Tryout)" }
+    ]
+  },
+  {
+    name: "Dustin",
+    gender: "m",
+    age: 56,
+    position: "Coach",
+    activePrograms: [
+      { name: "2019 ACC Academy", subtitle: "(Junior Academy Tryout)" },
+      { name: "2019 ACC Academy", subtitle: "(Junior Academy Tryout)" }
+    ]
+  }
+]);
+
+function loadData(page, itemsPerPage, sortBy) {
+  if (page == 1) {
+    return data.then(items => [items[0]]);
+  } else {
+    return data.then(items => [items[1]]);
+  }
+}
+
+function mapData(item) {
+  return {
+    id: item.name,
+    name: (
+      <div>
+        <div>{item.name}</div>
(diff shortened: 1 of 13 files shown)


src/js/components/Table/Paginated/helpers.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+export const usePagination = (
+  defaultItemsPerPage: number,
+  defaultCurrentPage: number
+) => {
+  const itemsPerPageStateHooks = React.useState(defaultItemsPerPage);
+  const currentPageStateHooks = React.useState(defaultCurrentPage);
+
+  return [itemsPerPageStateHooks, currentPageStateHooks];
+};
```
JordanPQuinn: This is slick

src/js/components/Table/Paginated/PaginatedTable.stories.tsx
```diff
@@ -0,0 +1,125 @@
...
+    programs: item.activePrograms.map((p, idx) => <div key={idx}>{p.name}</div>)
+  };
+}
+
+stories.add("Default", () => (
+  <PaginatedTable
+    columns={columns}
+    mapDataToRow={mapData}
+    loadData={loadData}
+    defaultItemsPerPage={1}
```
bradcypert: If you can set this to anything, how do we get the select list to match?

If you pass in a value that isn't one of the options, don't display the select?

src/js/components/Table/Paginated/PaginationSelect.tsx
```diff
@@ -0,0 +1,33 @@
+import { Select } from "../../Select";
+import * as PropTypes from "prop-types";
+import * as React from "react";
+
+const propTypes = {
+  options: PropTypes.arrayOf(PropTypes.number).isRequired,
+  setItemsPerPage: PropTypes.func.isRequired
+};
+
+const PaginationButtons: React.FunctionComponent<PropTypes.InferProps<
```
bradcypert: Rename to PaginationSelect

bradcypert:
I need to revisit the button logic as it feels a little wonky.

src/js/components/Table/Paginated/PaginationButtons.tsx
```diff
@@ -0,0 +1,91 @@
...
+        isDisabled={1 === currentPage}
+        onClick={() => setCurrentPage(1)}
+        type="button"
+      >
+        1
+      </Button>
+      {buttonLength > MAX_MIDDLE_BUTTONS && currentPage > MAX_MIDDLE_BUTTONS ? (
+        <Button className="Button" type="button">
+          ...
+        </Button>
```
amy-r: This may be more of a design question. Wouldn't we expect something to happen when we click this? I would thinking clicking this would display the hidden page numbers so that you can navigate to a specific page.

src/js/components/Table/Paginated/PaginationButtons.tsx
```diff
@@ -0,0 +1,91 @@
...
+          >
+            {index + page}
+          </Button>
+        );
+      })}
+      {buttonLength > MAX_MIDDLE_BUTTONS &&
+      currentPage + MAX_MIDDLE_BUTTONS <= totalItems ? (
+        <Button className="Button" type="button">
+          ...
+        </Button>
```
amy-r: Same with here. Another alternative would be making these spans instead of buttons.

src/js/components/Table/Paginated/PaginationButtons.tsx
bradcypert: I think the main point is to remain consistent with the button styling, but indicate that there are more pages than can be seen currently. We can definitely circle back and update this if there should be an action the user can take with these buttons.

src/js/components/Table/Paginated/PaginationButtons.tsx
bradcypert: I left a comment on the similar issue above explaining my thoughts.

src/js/components/Table/Paginated/PaginationButtons.tsx
amy-r: I think I'm misunderstanding. Why would we want something styled like a button if it doesn't do anything when you click it?

PR #266, opened 2020-03-04
Modify popups to be less opinionated

bradcypert:


src/js/components/Popup/PopupWrapper.tsx
```diff
@@ -0,0 +1,163 @@
+import * as React from "react";
+import * as PropTypes from "prop-types";
+
+interface State {
+  isPopupOpen: boolean;
+  isConfirmOpen: boolean;
+  selectedAction: any;
+}
+
+export default class PopUpWrapper extends React.Component<
```
dmccraw: Love this abstraction but could this just be called PopUp? Why add Wrapper? Is Wrapper a name we want to extend throughout teamsnap-ui to represent the smallest unit above the html?

PR #272, opened 2020-05-28
Add in disabled styles for feedback

bradcypert:
Trying to capture the style for this feedback (https://www.figma.com/proto/GNPNI5QlQy274h6nUAvePh/Checkout-Credits?node-id=73%3A145&scaling=min-zoom) when the child toggle is "off." I called it "Feedback--disabled" but Im happy to find a better name for it.

### DIFF
--- a/src/css/components/Feedback.scss
+++ b/src/css/components/Feedback.scss
@@ -106,7 +106,7 @@
     color: $cu-primary;
   }
   .Feedback-dismiss {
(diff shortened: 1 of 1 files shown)


saralohr:
That naming works for me. What's the orange dismiss for?

bradcypert:
@saralohr That'd be a control + z mistake! xD
Updated that to also be grey.

PR #283, opened 2020-09-21
Fix paging row number

dbhaasis:
Updates to fix a few things with paginated table

- Keep button count more consistent
- Update button styling for active vs disabled
- Hide buttons when only 1 page
- Allow ability to hide row select
- Select the row count for default page size


src/css/components/PaginationButton.scss
```diff
@@ -0,0 +1,16 @@
...
+  border-top-width: 2px;
+  border-bottom-width: 1px;
+  pointer-events: none;
+}
+
+.PaginateItemIsDisabled {
+  pointer-events: none;
+  color: #7a7a7a;
+  background: #f7f7f7;
+}
```
bradcypert, line 16: Can you add a blank line at the end here?

bradcypert (approved):

src/css/components/PaginationButton.scss
dbhaasis: blank line added

PR #284, opened 2020-09-24
Add in selectable row support for Paginated Table

bradcypert:
Additionally, change the type definition and callback for loadData to
use an object instead of an arbitary set of parameters.


src/js/components/Table/Paginated/PaginatedTable.tsx
```diff
@@ -76,11 +94,61 @@ const PaginatedTable: React.FunctionComponent<Props> = ({
...
+        } else {
+          setSelected([...selected, ele]);
+        }
+      }}}/>
+    </div>}));
+  }
+
+  // collisions overwrite
+  const bulkActionFuncsByLabel = bulkActions?.reduce((acc, action: BulkAction) => {
+    acc[action.label] = action.onSelected;
```
dbhaasis, line 128: Are the bulk actions the only way to access the selected items?

dbhaasis (approved):
Nice job on adding this feature.

src/js/components/Table/Paginated/PaginatedTable.tsx
bradcypert: Yeah, thee selected items state is maintained by the table. If we want to expose other ways to manage that, we can.

PR #285, opened 2020-09-28
Update service.yml

ryanwi:
@bradcypert does this serve from netlify or dugout? What's the base url?


ryanwi:
I see netlify activity on the PR fwiw

PR #290, opened 2020-09-29
Paginated table search

dbhaasis:
Add ability to further filter paginated results by a search box and also extra custom filter data.


bradcypert (approved):
The functions that are defined in BasicSearch, probably should use `React.useCallback`, but in this case, its probably not a huge deal. Basically, with function components, anything inside the function is going to run on each render. React.useCallback helps memoize these so we're not defining them time and time again.

For example, instead of:
```
function updateSearchField(e) {
    const searchTerm = e.target.value;

    setSearchValue(searchTerm);
  }
```

You could do
```
const updateSearchField = React.useCallback((e) => setSearchValue(e.target.value), []);
```
The empty dependency array in this example tells react that we dont need to redefine this function once its been defined and this component has mounted. If you do actually care, such as in the case of the `searchAction` function, you would do:
```
const searchAction = React.useCallback(() => handleSearch(searchValue), [searchValue]);
```

PR #323, opened 2020-10-10
Improve radio & checkbox list support in FieldWrapper

dobesv:
- Allow separate value field on each option to set input value
- Allow inputProps inside fieldProps to set props common to all checkbox and radio inputs in the options list
- Allow inputProps to be set in an option to set props on the input for that specific option
- Update storybook story


bradcypert:
👋 Hey Dobesv! Thanks for your contribution to TeamSnap UI!

Just a heads up, we've added a [License](https://github.com/teamsnap/teamsnap-ui/blob/master/LICENSE) since you've created this PR. If you can confirm that you're okay submitting your changes now that we have [an MIT license](https://github.com/teamsnap/teamsnap-ui/blob/master/LICENSE) applied, I'd be happy to help review and merge this PR.

Lemme know :)

dobesv:
The license is no problem, just trying to scratch my own itch.

dobesv:
BTW I have `console.log` in the storybook stories because the "action" call wasn't logging anything to the actions panel for some reason.  Feel free to edit the PR to remove those if you don't like and you're not having the same problem.

src/js/components/FieldWrapper/FieldWrapper.stories.tsx
```diff
@@ -63,7 +65,15 @@ stories.add("Checkbox", () => {
       field="checkbox"
       status={status}
       message={status === "error" ? errorMessage : null}
-      fieldProps={{ options: [{ label: "Check me" }, { label: "Check me" }] }}
+      fieldProps={{
+        inputProps: {
+          onChange: (e: ChangeEvent<HTMLInputElement>) => {
+            console.log(e.target.value + ' clicked');
+            action(e.target.value + ' clicked');
+          },
```
bradcypert, line 73: The reason your action isn't firing as expected is because the `action` function returns an event handler ([at least, in the version that we're currently on](https://github.com/storybookjs/storybook/tree/aeeafda2de3e97cd91e36c3dc7bdb0543d073429/addons/actions)). Ideally, the `action`'s responsne would be used in place of the onChange handler, but I like what you're doing with trying to differentiate. You should be able to do something like this instead:
```suggestion
          onChange: (e: ChangeEvent<HTMLInputElement>) => {
            action(e.target.value + ' clicked')(e);
          },
```
It might look a little weird, but you're just passing the synthetic event right to the action. I tested this locally and it worked for me :)

src/js/components/FieldWrapper/FieldWrapper.tsx
```diff
@@ -67,8 +67,16 @@ class FieldWrapper extends React.PureComponent<PropTypes.InferProps<typeof Field
...
       return (
         fieldProps.options &&
-        fieldProps.options.map(({ label }, i) => (
-          <FieldTag name={name} label={label} key={i} {...fieldProps} />
+        fieldProps.options.map(({ label, value, ...optionProps }, i) => (
+          <FieldTag
+            key={i}
+            {...fieldProps}
+            group={name}
+            name={name + String(i)}
```
bradcypert, line 75: Super small thing, we've started to prefer string interpolation in cases like this:
```suggestion
            name={`${name}${i}`}
```
Mind switching this over for us? :)

bradcypert (changes requested):
@dobesv Overall this looks great (thanks for helping us better support radio and checkboxes). I've got two small changes that I think we should make, then we'll be good to merge!

src/js/components/FieldWrapper/FieldWrapper.stories.tsx
dobesv: Oh right, I totally forgot about that.

bradcypert (approved):

bradcypert:
Thanks again for your contribution @dobesv! You can find your changes in v3.8.2 :)
https://www.npmjs.com/package/@teamsnap/teamsnap-ui/v/3.8.2

PR #332, opened 2020-10-19
Make it so pagination can exist outside of the table component

bradcypert:
This PR is intended to provide Pagination controls without necessarily requiring that you use a paginated table. Theres little to no reason that these components have to belong to the paginated table, so we're pulling them out to live on their own :)


wiseemily88 (approved):
Looks good to me!

PR #340, opened 2020-10-28
npm audit fix

ryanwi:
Patches security vulnerabilities

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12092,9 +12092,9 @@
       "integrity": "sha512-NuAESUOUMrlIXOfHKzD6bpPu3tYt3xvjNdRIQ+FeT0lNb4K8WR70CaDxhuNguS2XG+GjkyMwOzsN5ZktImfhLA=="
     },
     "object-path": {
-      "version": "0.9.2",
-      "resolved": "https://registry.npmjs.org/object-path/-/object-path-0.9.2.tgz",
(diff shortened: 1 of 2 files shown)


saralohr (approved):
Thank you! 🙌

bradcypert (approved):

bradcypert (commented):
why did this one go backwards? O.o

ryanwi:
> Do we want to cut a build for this right away or can we roll these changes up into the next one?

@bradcypert what's all involved in cutting a a build? If you think one is coming up, I am fine with rolling these changes up.

> why did the gulp dependency go backwards? O.o

Not sure about that, I was mostly letting `npm` do the work. Will check it out.

ryanwi:
> why did the gulp dependency go backwards? O.o

Ah, looks like the --force did it. I can remove that if it's breaking things. Which begs the question, how do I know if this breaks anything?

bradcypert:
@ryanwi Our CI currently just ensures that things build, but we also can check the storybook to ensure that our components still work `yarn storybook` should get you there :)

bradcypert:
FWIW: I think Gulp is only used for the CSS pipeline, so as long as the build succeeds, I think it should be fine.

ryanwi:
The results of `npm audit` do not change from the master branch to this branch after running audit fix, so I am removing those changes from the PR and just patching the one vulnerability with object-path.

bradcypert:
@ryanwi thanks! I'll get those bundled into a release soon!

PR #341, opened 2020-10-28
Move asset location

ryanwi:
Move asset location


saralohr (approved):
Don't know what's up with the build but font changes are 👍🏼

PR #358, opened 2020-11-03
Add in Avatar component

bradcypert:
Add in a React component to help with building "Avatars". These are needed in Organizations, and I imagine we're using circular images elsewhere, too.

Screen Shot 2020-11-03 at 2 14 41 PM

### DIFF
--- a/src/css/components/Avatar.scss
+++ b/src/css/components/Avatar.scss
@@ -2,23 +2,24 @@
 
 // We're working with images, not margins so we scale this up
 $sizes: (
-  "Xs": $su-xsmall * 5,
-  "Sm": $su-small * 5,
-  "Md": $su-medium * 5,
-  "Lg": $su-large * 5,
-  "Xl": $su-xlarge * 5,
-  "Xxl": $su-xxlarge * 5
+  "xs": $su-xsmall * 5,
+  "sm": $su-small * 5,
+  "md": $su-medium * 5,
+  "lg": $su-large * 5,
+  "xl": $su-xlarge * 5,
+  "xxl": $su-xxlarge * 5
 );
 
 .Avatar {
+
+  img {
+    border-radius: 50%;
+  }
+
   @each $name, $value in $sizes {
-    &-#{$name} {
+    &--#{$name} {
       width: $value;
       height: $value;
-
-      img {
-        border-radius: ($value / 2);
-      }
     }
   }
 }
--- a/src/css/components/Avatar.scss
+++ b/src/css/components/Avatar.scss
@@ -1,0 +1,24 @@
+@import "../config/index.scss";
+
+// We're working with images, not margins so we scale this up
+$sizes: (
+  "Xs": $su-xsmall * 5,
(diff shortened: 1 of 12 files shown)


bradcypert:
@saralohr I'm not sure how to get them added to this sidebar: https://teamsnap-ui-patterns.netlify.app/patterns/components/tabs.html

Any tips?

saralohr:
Yep - you have to bump the TS-UI submodule commit ref in the ui-patterns repo and then add an HMTL pattern for it. However we are having a deploy issue over there right now so you may want to hold on that for the moment. (sorry I need to look into that)

This is where the patterns live: https://github.com/teamsnap/ui-patterns/tree/master/src/patterns/components

src/css/components/Avatar.scss
```diff
@@ -0,0 +1,24 @@
...
+  "Sm": $su-small * 5,
+  "Md": $su-medium * 5,
+  "Lg": $su-large * 5,
+  "Xl": $su-xlarge * 5,
+  "Xxl": $su-xxlarge * 5
+);
+
+.Avatar {
+  @each $name, $value in $sizes {
+    &-#{$name} {
```
saralohr, line 15: According the [SUIT CSS naming conventions](https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md) that TS-UI follows, a modifier class would use double dash with the modifier in camel case.  So I think we could just keep these lower case, so it would be `Avatar Avatar--xs`. 

Although maybe we could add the image `border-radius` on the `Avatar` component so that it is still a freestanding element with the base style applied. And the modifier could add the size. So like:

```
.Avatar {
  img {border-radius: 50%}
}

@each $name, $value in $sizes {
  .Avatar--#{$name} {
    width: $value;
    height: $value;
  }
}

bradcypert: This comment is awesome! I learned so much from it!

1. I learned we use SUIT CSS naming conventions.
2. I learned that broder-radius can take a % value (I had no idea!)
3. I learned why we have the double dash in our class names sometimes!

Thanks @saralohr ! I've updated the code to better follow existing patterns!

saralohr: 💯 😄

saralohr (approved):
Awesome, thanks for the update. Looks great!

saralohr:
PS @bradcypert Maybe it's time to consider sunsetting ui-patterns in favor of Storybook so we can keep everything in one place.

bradcypert:
@saralohr The real benefit I see to UI patterns (that we'll have to figure out if/how we want to incorporate into storybook) is that you can see the plain HTML for implementing the components, so you dont _have_ to use React. For us, that's... not super important I guess since we're fully adopting React, but maybe for others its helpful. That being said, I'm not sure who all uses TeamSnap UI outside of us (and we dont have a great way to tell).

bradcypert:
@saralohr we can move the discussion to #360 so it doesn't get lost when I merge this in :)

PR #369, opened 2020-11-13
Buttons v4

bradcypert:
This PR modifies the styles for buttons to be more in line with the vision for V4. Additionally, as this is a major version, I've taken the liberty of fixing a few small issues with the button component. These are breaking changes and they resolve the issue captured in #370. Additionally, this PR converts the button component over to a function component to be in line with standards at TeamSnap and in the React community at large.

I encourage you to pull the PR and use `npm run storybook` to play with the buttons yourself :)


src/css/config/colors.scss
```diff
@@ -0,0 +1,81 @@
...
+  8: #993600,
+  9: #662400,
+  10: #331200,
+);
+@function orange($shade: 6) {
+  @return map-get($map: $oranges, $key: $shade);
+}
+
+$greens: (
+  1: #F3FFF8,
```
dbhaasis, line 68: Cool. I had not seen styles used this way before

dbhaasis (approved):
Haven't viewed the styles yet, but am looking forward to see new designs come alive

src/css/components/Button.scss
```diff
@@ -3,9 +3,9 @@
 // * @define default Button
 
 // Global default button colors that are themed
-$button-color:                    $cu-foreground !default;
-$button-border-color:                $cu-divider !default;
-$button-text-color:                  $cu-info !default;
+$button-color:                       white !default;
```
saralohr, line 6: If we want theming to work on v4 we'll want to abstract-ify colors similar to how we've done it currently. Themes will only apply to the team web app, but may be nice to consider earlier rather than later.

src/css/components/Button.scss
```diff
@@ -38,243 +36,96 @@ $Button-padding: 0 scaleGrid(2);
...
-      color: $Button-color;
-      opacity: $Button-disabled-opacity;
+      color: gray(5);
     }
   }
 }
 
+$permutations: (
+  "primary" blue(6) blue(5) blue(7),
+  "blue" blue(6) blue(5) blue(7),
```
saralohr, line 73: Is this to accommodate our existing Button--blue and Button--primary classes... or for theming? Since these are the same color now, ideally we wouldn't need both, and could just keep Button--primary. I know marketing uses Button--blue, but I don't know if we want to keep around a duplicate class for that reason.

src/css/components/Button.scss
```diff
@@ -38,243 +36,96 @@ $Button-padding: 0 scaleGrid(2);
...
+  "negative" red(6) red(5) red(7),
+  "orange" orange(6) orange(5) orange(7),
+  "green" green(6) green(5) green(7)
+);
+
+@each $name, $primary, $hover, $active in $permutations {
```
saralohr, line 79: Cool way of handling this!

src/css/components/Button.scss
```diff
@@ -38,243 +36,96 @@ $Button-padding: 0 scaleGrid(2);
...
     }
   }
 }
 
+$permutations: (
+  "primary" blue(6) blue(5) blue(7),
+  "blue" blue(6) blue(5) blue(7),
+  "negative" red(6) red(5) red(7),
+  "orange" orange(6) orange(5) orange(7),
+  "green" green(6) green(5) green(7)
```
saralohr, line 76: Not a big deal, but just thinking about consistency of naming. Do we want "green" to be positive if red is "negative"? The green button may have more use cases outside of the positive/negative context, so I could see that as an argument for keeping it "green".

src/css/components/Button.scss
bradcypert: this was to accommodate our existing classes. Happy to remove it as part of V4.

src/css/components/Button.scss
bradcypert: These match the existing button classes that we had for parity reasons. I am okay with changing these strictly to primary, negative, warning, and positive. I think that would technically be more clear with theming too.

saralohr: May want to loop @l3aconator and @andrewmeacham into this convo. Do we expect that we will use orange and green buttons outside of the context of positive/warning?

andrewmeacham: Green will be used to indicate "positive" states and actions, but also may be used as a secondary accent color for things like charts and graphs.

andrewmeacham: Orange will be used to surface warnings, but may also be used for things like badging and alerting the user to take an action.

saralohr: What about something like `--primary` (blue), `--secondary` (green),  `--tertiary` (orange), `--negative` (red) to accommodate broader use cases and future theming?

saralohr: Sorry for being wishy washy! But I'm on the fence on this one. I do still like primary, negative, warning and positive since those names have meaning associated with them and would promote consistency in theming and usage.

andrewmeacham: That works for me. I don't have a strong opinion on this, but your proposal makes sense for theming -- which I think we will want to expand on in future cycles.

src/css/components/Button.scss
```diff
@@ -3,31 +3,29 @@
...
 $Button-border-width: $border-width-small;
-$Button-border-bottom-width: $border-width-medium;
 $Button-disabled-opacity: 0.4;
 $Button-font-family: $tu-openSans-fontFamily;
 $Button-font-size: $tu-base-fontSize;
 $Button-font-weight: $tu-semibold-fontWeight;
 $Button-height: scaleGrid(4);
 $Button-line-height: $Button-height - 3;
-$Button-padding: 0 scaleGrid(2);
+$Button-padding: 0 scaleGrid(1);
```
saralohr, line 22: I think the default padding here is still scaleGrid(2).

src/css/components/Button.scss
```diff
@@ -38,243 +36,95 @@ $Button-padding: 0 scaleGrid(2);
...
+  "negative" color-negative(6) color-negative(5) color-negative(7),
+  "warning" color-warning(6) color-warning(5) color-warning(7),
+  "success" color-success(6) color-success(5) color-success(7)
+);
+
+@each $name, $color, $hover, $active in $permutations {
+  .Button--#{$name} {
+    background-color: $color;
+    border-color: $color;
+    color: $ts-white;
```
saralohr, line 82: I wonder if we want a colorNeutral(0) or something for white if we're not going to do the additional layer of abstraction with foreground, background, middleground, etc (which I support).

src/css/components/Button.scss
```diff
@@ -38,243 +36,95 @@ $Button-padding: 0 scaleGrid(2);
...
-}
-
-
 // * Modifier: Button Small
 
 $Button--small-font-size: $tu-small-fontSize;
 $Button--small-height: scaleGrid(3);
 $Button--small-line-height: $Button--small-height - 3;
-$Button--small-padding: 0 $su-small;
+$Button--small-padding: 0 $su-xsmall;
```
saralohr, line 104: I think we still want this to be $su-small (scaleGrid(1)). I'm also wondering if we prefer to use the size name or the scale function for these. In the past we've tried to use the size name everywhere, but I always forget what the names mean. Using the function may be more clear.

src/css/components/Button.scss
```diff
@@ -38,243 +36,95 @@ $Button-padding: 0 scaleGrid(2);
...
-}
-
-
 // * Modifier: Button Large
 
 $Button--large-font-size: $tu-large-fontSize;
 $Button--large-height: scaleGrid(5);
 $Button--large-line-height: $Button--large-height - 3;
-$Button--large-padding: 0 scaleGrid(3);
+$Button--large-padding: 0 scaleGrid(1);
```
saralohr, line 120: Large button padding looks like it stays at scaleGrid(2).

src/css/components/Button.scss
```diff
@@ -285,67 +135,12 @@ $Button--large-padding: 0 scaleGrid(3);
 $Button--huge-font-size: $tu-large-fontSize;
 $Button--huge-height: scaleGrid(10);
 $Button--huge-line-height: $Button--huge-height - 3;
-$Button--huge-padding: 0 scaleGrid(8);
+$Button--huge-padding: 0 scaleGrid(2);
 
 .Button--huge {
```
saralohr, line 140: I think we can ditch this guy altogether. @l3aconator are we still using this behemoth?

src/css/components/Button.scss
bradcypert: Thank you. Is there a way in figma to see that? I tried to find margins or padding and I'm not seeing anything there.

src/css/components/Button.scss
bradcypert: Thoughts on color-neutral(1)? For TeamSnap's colors its #FBFBFB. Not sure if thats close enough, but my poor eyesight cant tell the difference.

src/css/components/Button.scss
bradcypert: The functions feel more strict to me, so I think that approach resonates best with me. I'm fairly un-opinionated on this piece in particular though.

.storybook/config.js
```diff
@@ -4,6 +4,7 @@ import { withInfo } from "@storybook/addon-info";
 
 // Import css
 import "../src/css/teamsnap-ui.scss";
+// import "../src/css/themes/custom-aquablue.scss";
```
bradcypert, line 7: This will be used for development to ensure that theming can still work with our changes.

src/css/components/Button.scss
saralohr: Cool, let's go with function names then.

src/css/components/Button.scss
saralohr: The padding isn't explicitly stated, but knowing that we want to keep the 8px grid, and looking at the buttons here: https://www.figma.com/proto/1c8omRfECeqirRZAutk0hK/New-UI?node-id=97%3A48&scaling=min-zoom I can see that default and large have roughly 16px of padding and the small button has about half that.

src/css/components/Button.scss
saralohr: Bacon is going to move this into the marketing repo if they want to keep using it, so we can remove it for now from TS-UI.

src/css/components/Button.scss
saralohr: I can barely tell the difference visually, but it does drop it below the AAA standard for accessibility, which is 7.0. When the text is white it's 7.15, and at color-neutral(1) it's 6.91. I think ideally we would aim for the AAA standard when possible. I also don't know if we have any other plans for color-neutral(1) such that we could make it full white instead of fbfbfb. Any reason to not add the additional number 0 for white?

Looping in @andrewmeacham for his thoughts.

src/js/components/Button/Button.stories.tsx
```diff
@@ -66,97 +66,53 @@ stories.add("TeamSnap Colors", () => {
   );
 });
 
-stories.add("Status Colors", () => {
```
saralohr, line 69: Just looking in Storybook and we have two stories for colors: Themed colors and TeamSnap colors. I think we can consolidate this to just Colors since we will be making a theme for TeamSnap, and show all the buttons colors there.

saralohr: For that matter, I'm not sure we need the Default story either. That could be included in the colors story:
default, primary, positive, warning, negative

src/css/components/Button.scss
bradcypert: Gonna go ahead and merge this in, but we can circle back to this at a later date if needed. This should be very easy to adapt if we decide we need a new color for it.

saralohr:
Thanks @bradcypert it looks great! 🏅 Thanks for putting up with all my comments 😄

PR #391, opened 2020-12-01
Left Navigation

bradcypert:
Will have to circle back to this for more treatment on the Nav Header piece when we figure out exactly what we want to do there, but I think the rest is pretty solid and good to go! Will tag a few people with react experience to get their opinions on the way the component works.

@dbhaasis @stevencwarren @wiseemily88, can I get any feedback on the demo implementation in the storybook? Does that seem like an okay way to use this component? Is it confusing? Would you expect it to be configured and used differently?

### DIFF
--- a/src/css/components/Avatar.scss
+++ b/src/css/components/Avatar.scss
@@ -24,4 +24,9 @@ $sizes: (
       height: $value;
     }
   }
+
+  &--fill {
+    display: flex;
+    width: 100%;
+  }
 }
--- a/src/css/components/Nav.scss
+++ b/src/css/components/Nav.scss
@@ -23,7 +23,7 @@
       padding: 16px;
     }
 
-    .Nav-collapse {
+    .Nav-footer {
       .Icon {
         transform: rotate(180deg);
       }
@@ -37,7 +37,7 @@
   font-size: 1.4rem;
 }
 
-.Nav-collapse {
+.Nav-footer {
   border-top: 1px solid color-neutral(5);
   cursor: pointer;
   padding: 12px 24px;
--- a/src/css/components/Nav.scss
+++ b/src/css/components/Nav.scss
@@ -4,7 +4,7 @@
   display: flex;
   flex-direction: column;
 
-  a, .nav-item {
+  a, .Nav-item {
     color: color-neutral(9);
   }
 
@@ -15,57 +15,57 @@
 
   &.is-collapsed {
     max-width: 64px;
-    .nav-item-title {
+    .Nav-item-title {
       display: none !important;
     }
 
-    .nav-header {
+    .Nav-header {
       padding: 16px;
     }
 
-    .nav-collapse {
+    .Nav-collapse {
       .Icon {
         transform: rotate(180deg);
       }
     }
   }
+}
 
-  .nav-header {
-    border-bottom: 1px solid color-neutral(5);
-    padding: 16px;
-    font-size: 1.4rem;
+.Nav-header {
+  border-bottom: 1px solid color-neutral(5);
+  padding: 16px;
+  font-size: 1.4rem;
+}
+
+.Nav-collapse {
+  border-top: 1px solid color-neutral(5);
+  cursor: pointer;
+  padding: 12px 24px;
+  .Icon {
+    transition: transform 0.2s ease-in-out;
+  }
+}
+
+.Nav-item {
+  background-color: color-neutral(1);
+  border-radius: 5px;
+  cursor: pointer;
+  height: 42px;
+  list-style: none;
+  margin: 12px;
+  padding: 12px;
+  user-select: none;
+
+  &:hover {
+    background-color: color-neutral(3);
   }
 
-  .nav-collapse {
-    border-top: 1px solid color-neutral(5);
-    cursor: pointer;
-    padding: 12px 24px;
-    .Icon {
-      transition: transform 0.2s ease-in-out;
-    }
+  &:active {
+    background-color: color-neutral(4);
   }
 
-  li {
-    background-color: color-neutral(1);
-    border-radius: 5px;
-    cursor: pointer;
-    height: 40px;
-    list-style: none;
-    margin: 12px;
-    padding: 12px;
-    user-select: none;
-
-    &:hover {
-      background-color: color-neutral(3);
-    }
-
-    &:active {
-      background-color: color-neutral(4);
-    }
-
-    &.is-active {
-      background-color: color-primary(6);
-      color:color-neutral(1);
-    }
+  &.is-active {
+    background-color: color-primary(6);
+    color:color-neutral(1);
   }
 }
--- a/src/css/components/Nav.scss
+++ b/src/css/components/Nav.scss
@@ -16,11 +16,13 @@
 
   &.is-collapsed {
     max-width: scaleGrid(8);
+    min-width: scaleGrid(8);
     .Nav-itemTitle {
       display: none !important;
     }
 
     .Nav-header {
+      margin: 0 auto;
       padding: scaleGrid(2);
     }
 
@@ -43,6 +45,11 @@
 .Nav-body {
   flex: 1 1 0% !important;
   flex-basis: 0% !important;
+}
+
+.Nav-headerIcon {
+  width: scaleGrid(3);
+  display: block;
 }
 
 .Nav-footer {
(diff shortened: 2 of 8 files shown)


src/css/components/Nav.scss
```diff
@@ -0,0 +1,71 @@
...
+    &:active {
+      background-color: color-neutral(4);
+    }
+
+    &.is-active {
+      background-color: color-primary(6);
+      color:color-neutral(1);
+    }
+  }
+}
```
saralohr, line 71: Could the class structure could be cleaned up a little so the output is more like:
```
<div class="Nav">
  <div class="Nav-header">
    <div class="Nav-headerIcon" /> ...etc
  <div class="Nav-body">
    <div class="Nav-item">
    <div class="Nav-item">
  <div class="Nav-footer">
</div>
```

src/css/components/Nav.scss
```diff
@@ -0,0 +1,71 @@
...
+  .nav-collapse {
+    border-top: 1px solid color-neutral(5);
+    cursor: pointer;
+    padding: 12px 24px;
+    .Icon {
+      transition: transform 0.2s ease-in-out;
+    }
+  }
+
+  li {
```
saralohr, line 48: Can this be `.Nav-item` instead of an li?

src/css/components/Nav.scss
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    .nav-collapse {
+      .Icon {
+        transform: rotate(180deg);
+      }
+    }
+  }
+
+  .nav-header {
```
saralohr, line 33: Ideally we'd unnest everything from .Nav that doesn't need to be descendent or attached (like the `is-collapsed` styles and pseudo elements). The CSS will render (ever so infinitessimally) faster if it just has the one element to parse. Also Can these inherit the Title case `Nav` pretty please, eg `Nav-header` 🙏

src/js/components/Nav/Nav.tsx
```diff
@@ -0,0 +1,132 @@
...
+  );
+
+  return (
+    <nav className={cname} style={style} {...otherProps}>
+      {headerItems && headerItems.length ? (
+        <div className="nav-header u-textSemiBold u-flex">
+          <div
+            className={`${
+              !isCollapsed ? "u-size1of8" : ""
+            } nav-header-icon u-flex u-flexAlignItemsCenter u-flexJustifyCenter`}
```
saralohr, line 95: I don't think you need the `u-flex u-flexAlignItemsCenter u-flexJustifyCenter` on every child of the parent `nav-header u-flex` container. Adding `u-flexAlignItemsCenter` to the parent (nav-header) should do the trick.

However... if the nav header will always display flex, may be cleaner to move those to styles on the component:
```
.Nav-header {
  display: flex;
  align-items: center;
}
```

src/js/components/Nav/Nav.stories.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { storiesOf } from "@storybook/react";
+import { text } from "@storybook/addon-knobs/react";
+import Nav from "./Nav";
+const stories = storiesOf("Nav", module);
+
+stories.add("Default", () => (
+  <div
+    className="u-flex"
+    style={{ backgroundColor: "#FFEFA5", minHeight: "800px" }}
+  >
```
saralohr, line 11: Been playing around with CSS Grid in some personal projects, and it is incredibly cool and powerful, and finally has enough browser support to consider adopting. Flex is still awesome for many things, but this layout may lend itself really well to a Grid flow (the Nav and Main sections). Something worth considering for future iterations.

src/css/components/Nav.scss
bradcypert: I think I've cleaned it up so that its more aligned with what you're asking, but I definitely would appreciate a second pass on reviewing it again.

src/js/components/Nav/Nav.tsx
bradcypert: I went ahead and pulled that out to Nav-header. It feels a lot cleaner, I think I'm just getting caught up trying to balance adding additionally CSS vs using existing classes. Ideally we can use this opportunity to shrink the CSS file, too, but I might need some guidance on that (or some help planning that out).

src/js/components/Nav/Nav.tsx
```diff
@@ -0,0 +1,139 @@
...
+              src={headerItems[0].image}
+              size="fill"
+            />
+          </div>
+          <div className="u-sizeFill Nav-item-title u-spaceLeftSm">
+            <span className="Nav-item-title u-textEllipsis">
+              {headerItems[0].title}
+            </span>
+          </div>
+          <div className="u-size1of8 Nav-item-title">
```
dbhaasis, line 114: So, if we are always showing down arrow on header, how are we controlling the sub menu when you click on down?

src/js/components/Nav/Nav.tsx
```diff
@@ -0,0 +1,139 @@
...
+            <Icon mods="u-fontSizeMd" name="down" />
+          </div>
+        </div>
+      ) : null}
+      <div className="Nav-body u-sizeFill">{children}</div>
+      <div className="Nav-footer" onClick={() => setCollapsed(!isCollapsed)}>
+        <Icon name="left" />{" "}
+        <span className="Nav-item-title">Collapse Menu</span>
```
dbhaasis, line 122: Love the collapse

src/js/components/Nav/Nav.tsx
bradcypert: We aren't yet. Sorry, this was posted in the project-specific slack channel, but we're going to circle back to this once we know exactly what we want it to do.

src/js/components/Nav/Nav.tsx
saralohr: The utility classes are useful to modify an existing component for one-off behavior changes, or for making layouts without having to create a whole new component for that one layout.

I don't typically think of them as a path to shrinking component CSS. Ideally component styles are completely encapsulated within the component classes, so that the component doesn't need any additional classes to display correctly. Our React components nullify that need a tad, but it's still best practice for CSS composition, and for if these components are ever used outside of React.

saralohr: The utility styles also compress very tightly due to their repetition. But I think we can probably sort through all of them to make sure we need/want them all.

src/css/components/Avatar.scss
```diff
@@ -24,4 +24,9 @@ $sizes: (
       height: $value;
     }
   }
+
+  &--fill {
```
saralohr, line 28: Don't think we actually need this modifier. In the header, the image can be centered vertically by setting it to display block (will remove the extra space below it). I also think we probably want this to have a fixed width so it doesn't scale in size with the screen.

```
.Nav-headerIcon {
  width: scaleGrid(3);
  display: block;
}

<div class="Nav-header">
  <img class="Nav-headerIcon"/>
</div>
```

src/css/components/Nav.scss
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  .Icon {
+    height: 1.2em;
+    width: 1.2em;
+  }
+
+  &.is-collapsed {
+    max-width: 64px;
+    .Nav-item-title {
```
saralohr, line 18: Nit: could this be `.Nav-itemTitle` instead?

src/js/components/Nav/Nav.stories.tsx
```diff
@@ -0,0 +1,81 @@
...
+    <Nav
+      mods="u-size2of12"
```
dbhaasis, line 13: I think fine to use mods for size of menu as long as it handles not having a mod for size.

dbhaasis (approved):
Seems pretty easy to set up.

src/js/components/Nav/Nav.stories.tsx
bradcypert: If you do omit the mods, it is as big as it's parent container, which should work out fine.

src/css/components/Avatar.scss
bradcypert: This modifier on the avatar component will also solve an issue that we're seeing in org, but I've added the CSS that you've recommended and got it behaving as expected, as well.

saralohr (approved):
Nice, thanks for humoring all of my nit-picky CSS changes. Looks great!

bradcypert:
@saralohr no worries! It helps me make notes and get a better understanding of how we do CSS here. :)

PR #393, opened 2020-12-01
Make popup more flexible

saralohr:
Make the popup confirm component more flexible by supporting extra classes, variable button text and adding direction. I added defaults for all the old static properties so it shouldn't impact components in the wild.

popup


wiseemily88 (approved):
nice!

bradcypert:
Great work! :D

PR #399, opened 2020-12-03
V4 inputs

bradcypert:
I think this is good for an initial review. Im sure the CSS could use some guidance and I know the box-shadows aren't 100% correct.

On the TypeScript side, I decided to simplify the interface for working with form fields. Instead of manually managing FieldWrappers FieldMessages and more, I decided to take the approach where the user should just facade through a Field. A field is still composed of those other components, so they're still available, but the recommended use case is using `<Field>` instead.

On that note, I added a small deprecated note to the storybooked components for form fields that ARENT the `<Field>` component.

I added a new folder to hold shared types (prop types, enums, etc) that are used across components. The goal here is to easily support (from the TypeScript side) any changes to sizes, statuses, etc in one easy to maintain place.

Please feel free to review the entire PR, but I could particularly use thoughts around:

1. How do I make this SCSS not terrible?
2. Can I simplify the Component Interface at all
3. Is this confusing? If so, how can I make it more clear?

Checking out the branch and running `npm run storybook` is encouraged 😄


src/css/components/FieldGroup.scss
```diff
@@ -51,16 +55,20 @@ $FieldGroup-font-size: $tu-base-fontSize;
 
 .FieldGroup.is-notValid {
   .Input, .Checkbox-label:before, .SelectBox-options, textarea, .Checkbox--pill .Checkbox-label {
-    border-color: $cu-negative;
+    border-color: color-negative(5);
+  }
+
+  .Input:focus {
+    box-shadow: 0px 0px 15px -5px color-negative(3);
   }
```
bradcypert, line 63: All of the box shadows are probably too big.

bradcypert:
Just walked through the box shadows with @andrewmeacham and he's onboard with how they look currently.

andrewmeacham (approved):

src/js/components/Field/Field.stories.tsx
```diff
@@ -0,0 +1,208 @@
...
+      placeholder={placeholder}
+      rightIcon={<Icon name="search" />}
+      showStatus
+      size={size}
+      status={status}
+    />
+  );
+});
+
+stories.add("Clear Icon", () => {
```
dbhaasis, line 176: Can you specify a clear icon or control what clears the input? It seems a bit strange to clear by clicking the status icon.

bradcypert: Sorry David, we found that icon moments ago and I hadnt pushed it up yet. All of these icons will eventually be replaced by the new icons, so we're trying not to get too hung up on them at the moment.

saralohr:
Thanks @bradcypert! I'll take a closer look on Monday, but didn't see anything too crazy in the CSS on first glance. 

The `FieldWrapper` component is a beast. Excited to pull this down and check out the new approach.

src/js/components/Input/Input.tsx
```diff
@@ -15,63 +15,96 @@
...
+    'u-flex',
+    mods
+  );
 
-    return (
-      <div className={inputClasses} style={style} {...otherProps}>
+  return (
+    <div className={inputClasses} style={style} {...otherProps}>
+      {leftIcon && <div className="icon-Left InputGroup-icon">{leftIcon}</div>}
+      <div className="u-sizeFill">
```
saralohr, line 79: Don't think you need this wrapper or the parent `u-flex` class on `InputGroup`. `Input` already has style `display: block; width: 100%` which will fill the space.

src/js/components/Input/Input.tsx
```diff
@@ -15,63 +15,96 @@
...
+    rightIcon && `InputGroup--rightIcon`,
+    (showStatus || showClear) && "InputGroup--auxIcon",
+    'u-flex',
+    mods
+  );
 
-    return (
-      <div className={inputClasses} style={style} {...otherProps}>
+  return (
+    <div className={inputClasses} style={style} {...otherProps}>
```
saralohr, line 77: I noticed here that when a different size is selected, the `InputGroup` wrapper is getting `Input--large/small` in addition to being applied to the nested `Input` element. It only needs to be on the `Input` element (not that it's harming anything here, but just isn't necessary and is kinda weird to have a modifier class on a different component name).

src/css/components/InputGroup.scss
```diff
@@ -45,22 +46,40 @@ $form-input-height: $su-xlarge !default;
 
 .InputGroup--leftIcon {
   .Input {
-    padding-left: scaleGrid(5);
+    padding-left: scaleGrid(4);
   }
 
-  .InputGroup-icon {
+  .icon-Left {
```
saralohr, line 52: Since this is a modifier of  `.InputGroup-icon`,  I would probably do something like `.InputGroup-icon--left` and same for the others: `.InputGroup-icon--right` and `.InputGroup-icon--aux`.

src/css/components/InputGroup.scss
```diff
@@ -45,22 +46,40 @@ $form-input-height: $su-xlarge !default;
...
+    right: 0;
+  }
+}
+
+.InputGroup--auxIcon {
+  .Input {
+    padding-right: scaleGrid(4);
+  }
+
+  .icon-Aux {
```
saralohr, line 82: When this is used for `showClear` I think it could use some styles to indicate it is interactive. And maybe it gets another modifier class added for the `showClear` state so it doesn't impact the status colors. Something like:

```
.InputGroup-icon--aux { // same as current
  right: 0;    
}

.InputGroup-icon--clear {
  cursor: pointer;
  transition: color ease 0.2s;
  &:hover {
    color: color-neutral(8);
  }
}
```

src/js/components/Field/Field.tsx
```diff
@@ -0,0 +1,90 @@
...
+import * as React from "react";
+import * as PropTypes from "prop-types";
+import { FieldGroup } from "../FieldGroup";
+import { FieldLabel } from "../FieldLabel";
+import { FieldMessage } from "../FieldMessage";
+import { Input } from "../Input";
+import { Status, Size } from "../../types";
+
+const fieldPropTypes = {
+  type: PropTypes.oneOf(["toggle", "select", "input"]).isRequired,
```
saralohr, line 24: How does this work with Checkbox/Radio... or is that coming later?

src/js/components/Field/Field.tsx
```diff
@@ -0,0 +1,90 @@
...
+  rightIcon: PropTypes.node,
+  size: Size.PropType,
+  showStatus: PropTypes.bool,
+  showClear: PropTypes.bool,
+  onClearClicked: PropTypes.func,
+  isDisabled: PropTypes.bool,
+};
+
+type FieldType = React.FunctionComponent<
+  PropTypes.InferProps<typeof fieldPropTypes>
```
saralohr, line 43: Is this because it takes the children props as arguments?

Screen Shot 2020-12-07 at 12 38 55 PM

src/js/components/Input/Input.tsx
```diff
@@ -15,63 +15,96 @@
...
+          placeholder={placeholder}
+          className={`Input ${size && `Input--${size}`} ${isDisabled && `Input--isDisabled`}`}
           {...inputProps}
         />
-        {children && <span className="InputGroup-icon">{children}</span>}
       </div>
-    );
-  }
+      {rightIcon && <div className="icon-Right InputGroup-icon">{rightIcon}</div>}
+      {(showStatus || showClear) && <div className="icon-Aux InputGroup-icon">
```
saralohr, line 91: This is cool!

saralohr (commented):
Nice work - this seems much more intuitive to work with than `FieldWrapper`!

src/js/components/Field/Field.tsx
bradcypert: Coming soon!™

src/css/components/InputGroup.scss
bradcypert: Great suggestion! This looks so much better!

src/js/components/Field/Field.tsx
bradcypert: Nope. I forgot to attach them to the function component. :) Fixed!

src/css/components/Checkbox.scss
```diff
@@ -161,13 +161,23 @@ $Checkbox-bg-color: $cu-foreground !default;
...
   // SVG check icon
   border: none;
 }
 
+// Active state
+.Checkbox--indeterminate .Checkbox-input:checked + .Checkbox-label:before {
+  background-image: url("data:image/svg+xml,%3Csvg width='4' height='4' viewBox='0 0 8 8' fill='none' xmlns='http://www.w3.org/2000/svg'%3E%3Crect x='1.5' y='1.5' width='5' height='5' rx='1.5' fill='%230F52BA' stroke='%230F52BA'/%3E%3C/svg%3E%0A");
+  background-color: color-neutral(1);
+  border: solid $border-width-small $cu-divider;
+}
```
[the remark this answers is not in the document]
bradcypert: The primary use case we have right this moment is that we have a table with selectable rows. When some of the rows are selected, the select all checkbox should hit that indeterminate state. As for the other use cases, I'm not sure of how many we'll have. @andrewmeacham might have a better idea as to their use cases.

src/js/components/Field/Field.tsx
```diff
@@ -19,72 +26,113 @@ import { FieldLabel } from "../FieldLabel";
 import { FieldMessage } from "../FieldMessage";
 import { Input } from "../Input";
 import { Status, Size } from "../../types";
+import { Checkbox } from "../Checkbox";
+import { CheckboxState, CheckboxStates } from "../../types";
+
+const checkboxShape = PropTypes.shape({
```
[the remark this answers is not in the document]
bradcypert: I dont think you'll want more than one checkbox in a field. I think the recommended approach will be to use multiple fields. I can add a storybook example for that and make sure it works as expected.

src/css/components/Checkbox.scss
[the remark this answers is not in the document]
saralohr: Ah, cool. That makes sense!

src/js/components/Field/Field.tsx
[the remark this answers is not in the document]
saralohr: I was thinking of a "Check all that apply" type scenario, where you would want one message and one label per group:
Screen Shot 2020-12-08 at 8 51 26 AM

And would want this for sure for radios.

src/css/components/Checkbox.scss
```diff
@@ -169,13 +169,16 @@ $Checkbox-bg-color: $cu-foreground !default;
   background-image: url("data:image/svg+xml; utf8, %3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 960 960'%3E%3Cpath fill='white' d='M915 178.9c8 6.7 12.3 15.5 13 26.5s-2.3 21.2-9 30.5c-252.7 360.7-389.3 555.3-410 584-21.3 28.7-48.8 43.2-82.5 43.5S365 849.2 343 819.9l-278-391c-6.7-9.3-9.7-19.7-9-31s5-20.3 13-27c34-29.3 72-55.7 114-79 8.7-4.7 18.3-5.3 29-2s19.3 9.7 26 19l188 264 321-456c6.7-9.3 15.2-15.5 25.5-18.5s20.2-2.2 29.5 2.5c44 24 81.7 50 113 78z'%3E%3C/path%3E%3C/svg%3E");
   // SVG check icon
   border: none;
+  transition: none
 }
 
 // Indeterminate state
 .Checkbox--indeterminate .Checkbox-input:checked + .Checkbox-label:before {
   background-image: url("data:image/svg+xml,%3Csvg width='4' height='4' viewBox='0 0 8 8' fill='none' xmlns='http://www.w3.org/2000/svg'%3E%3Crect x='1.5' y='1.5' width='5' height='5' rx='1.5' fill='%230F52BA' stroke='%230F52BA'/%3E%3C/svg%3E%0A");
```
saralohr, line 177: Sorry should have called this out but I updated these attributes in the SVG code too:
`x='2' y='2' width='4' height='4' rx='1'`

bradcypert: @saralohr sorry, which SVG? The Active one?

src/js/components/Field/Field.stories.tsx
```diff
@@ -192,17 +207,92 @@ stories.add("Clear Icon", () => {
...
+        type="checkbox"
+        label="Accept Terms of Service"
+        formFieldProps={{
+          text: "Click me!",
+          checked: value,
+          onClick: () => {
+            setValue(!value);
+          },
+        }}
+        name="Sample"
```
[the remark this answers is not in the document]
bradcypert: This should be fixed now. My bad!

src/js/components/Field/Field.tsx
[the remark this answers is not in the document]
bradcypert: Pushed up an example that uses multiple checkboxes and manual Field.Label and Field.Caption usage to indicate how this could be accomplished. We may revisit this once we tackle radios, as we may find a better way to handle this while building them out.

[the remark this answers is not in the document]
saralohr: Cool, that seems fine for now. I imagine we'll want it integrated in the future to handle required fields, error messages, etc, but we can tackle that when we get to it.

src/css/components/Checkbox.scss
saralohr: Oh sorry, the indeterminate one. It just makes the inner box a little smaller since we are making the background size 100%.

saralohr (approved):
🎉

src/js/components/Field/Field.tsx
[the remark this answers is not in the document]
bradcypert: I really want to be cautious there as I dont think our component library should dictate how a user does form validation. My goal, especially for form fields, is to expose whats needed so that people can wire up whatever they want to work with these fields.

PR #429, opened 2021-01-08
React 17

saralohr:
Update TS-UI to React 17 and fix breaking changes for event delegation:

https://reactjs.org/blog/2020/08/10/react-v17-rc.html#changes-to-event-delegation

I didn't find any other issues. `onFocus` and `onBlur` have changed, but I didn't find those in our JS source. May be something to keep in mind for teamsnap-spas projects though.

@bradcypert would love your 👀 on this in case there is anything else we should look out for.


wiseemily88 (approved):
🔥 looks good to me!

bradcypert (approved):

PR #451, opened 2021-02-01
add in toolbar for alphav4

bradcypert:
Icons are temporary until we replace Pika with whatever icon set we're using for v4.

Screen Shot 2021-02-01 at 3 56 11 PM

Screen Shot 2021-02-01 at 3 58 10 PM

### DIFF
--- a/src/css/utils/flex.scss
+++ b/src/css/utils/flex.scss
@@ -73,6 +73,10 @@
   justify-content: space-around !important;
 }
 
+.u-flexJustifyEvenly {
+  justify-content: space-evenly !important;
+}
+
 
 // * Align items in the cross axis of the current line of the flex container
 // * Similar to `justify-content` but in the perpendicular direction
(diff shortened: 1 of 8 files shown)


src/css/utils/flex.scss
```diff
@@ -73,6 +73,10 @@
   justify-content: space-around !important;
 }
 
+.u-flexJustifyEvenly {
+  justify-content: space-evenly !important;
```
saralohr, line 77: I don't know how, but I completely missed the existence of this flex property. It's weird that SUIT CSS doesn't include it either 🤷‍♀️

saralohr (approved):
Looks good!

src/css/utils/flex.scss
bradcypert: No worries :) I tried a few examples of what I thought it would be and when none of them worked, I realized we were missing it ! D:

jlantweiler:
👍

src/css/utils/flex.scss
saralohr: Nice, TIL!

PR #538, opened 2021-03-30
Adding custom filters for paginated table component

BillyFigueroa:
This PR is related to...

- [https://teamsnap.atlassian.net/browse/SO-4319](url)

It is adding the ability to use a previously created attribute in `PaginatedTable` to pass a custom filter component

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,7 +91,6 @@
     "lodash": "^4.17.15"
   },
   "peerDependencies": {
-    "react": "^17.0.1",
-    "react-dom": "^17.0.1"
+    "react": "^17.0.1"
   }
 }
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -91,6 +91,7 @@
     "lodash": "^4.17.15"
   },
   "peerDependencies": {
-    "react": "^17.0.0"
+    "react": "^17.0.1",
+    "react-dom": "^17.0.1"
```
bradcypert, line 95: Billy, I did some more research and I was wrong about needing to add react-dom here. We'll want to remove it from the peer dependencies list please. Sorry :(

bradcypert (approved):

PR #557, opened 2021-04-05
Allow pagination to be displayed below the paginated table

bradcypert:
Screen Shot 2021-04-05 at 3 22 49 PM

### DIFF
--- a/src/js/utils/placement.ts
+++ b/src/js/utils/placement.ts
@@ -1,0 +1,4 @@
+export enum Placement {
+  Bottom,
+  Top
+}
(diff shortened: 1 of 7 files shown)


src/js/utils/placement.ts
```diff
@@ -0,0 +1,4 @@
+export enum Placement {
```
BillyFigueroa, line 1: Is utils the best place for this? Do we have a config/objects type of folder location or something similar?

dbhaasis (approved):
Code changes look good.

src/js/utils/placement.ts
bradcypert: Funny you should mention that! @bradjteamsnap and I were just talking about how in the V4 branch, we have a `types` folder.  Im okay with keeping it in utils on this branch since there isn't already a home, but in V4 it will live in `types` and, if we keep v3 around much longer, I am happy to move it to a more suitable home once we have one :)

BillyFigueroa: Sounds good

PR #589, opened 2021-04-16
So 4418 create pill components

BillyFigueroa:
Creating a base pill component as well as more complicated alternative components that use the base pill component


bradjteamsnap (approved):
lgtm

PR #593, opened 2021-04-21
So 4435 create recipients selectable list

BillyFigueroa:
Creating various components associated with the recipient list components

- Tag
- ListToggle
- PanelList

These components will make up a large majority of the new email page

Screen Shot 2021-04-21 at 12 31 37 AM

### DIFF
--- a/src/js/components/ToggleCheckboxList/ToggleCheckboxList.tsx
+++ b/src/js/components/ToggleCheckboxList/ToggleCheckboxList.tsx
@@ -9,7 +9,7 @@ import { ListToggle } from '../ListToggle';
 
 type Props = {};
 
-const PanelList:FunctionComponent<Props> = () => {
+const ToggleCheckboxList:FunctionComponent<Props> = () => {
   const [activeRows, setActiveRows] = useState([]);
   const [headerStatus, setHeaderStatus] = useState({});
 
@@ -204,4 +204,4 @@ const PanelList:FunctionComponent<Props> = () => {
   );
 }
 
-export default PanelList;
+export default ToggleCheckboxList;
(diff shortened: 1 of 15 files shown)


src/js/components/ToggleCheckboxList/ToggleCheckboxList.tsx
```diff
@@ -0,0 +1,207 @@
...
+    // Row already exists, unselect it
+    if (activeRows.includes(division)) {
+      newActiveList = activeRows.filter(item => item != division);
+
+      activeCount = headerStatus[productName] ?
+        headerStatus[productName]['activeCount'] - 1 :
+        headerStatus[productName]['activeCount'];
+
+      if (activeCount > 0 && activeCount < children.length) {
+        newStatus = 'indeterminate';
```
bradjteamsnap, line 98: nit: use constants here for the state values: 
ex `import { CheckboxStates } from "../../../types";`
`CheckboxStates.INDETERMINATE`

src/js/components/ToggleCheckboxList/ToggleCheckboxList.tsx
```diff
@@ -0,0 +1,207 @@
...
+import { Field } from '../Field';
+import { ListToggle } from '../ListToggle';
+
+type Props = {};
+
+const ToggleCheckboxList:FunctionComponent<Props> = () => {
+  const [activeRows, setActiveRows] = useState([]);
+  const [headerStatus, setHeaderStatus] = useState({});
+
+  const programData = [{
```
bradjteamsnap, line 16: nit: I know you have plans to clean up after the data names are more final, but let's not define this here. Maybe for now define this test data in the story and pass it in as a prop.

bradjteamsnap (approved):
left a few non-blocking nits since this is v4 work in progress. otherwise lgtm!

PR #621, opened 2021-05-06
V4 remove classic icons

andrewmeacham:
- Replaced pika in SelectBox component
- Removed pika font and references

Screen Shot 2021-05-06 at 11 32 17 AM

Note: material icon font will be moved to classic.


saralohr (approved):
Love it! Thanks for doing this!

PR #623, opened 2021-05-06
Adding a select all function to component

BillyFigueroa:
Adding a select all function to component so we can reference it with a React Ref and be able to use a 'select all' button call from an outside component


bradjteamsnap (approved):
looks ok. The `selectAllRows` could maybe be memoized since it is a bit high on complexity. but prob ok to deal with it only if it becomes an issue.

PR #633, opened 2021-05-14
Modify buttons to have a minimum height

bradcypert:
Modify button css to have a minimum height instead of a fixed height. This should help fix issues where a lot of text is shoved in a small button and bleeds over the border boundary by allowing the button to grow.

### DIFF
--- a/src/css/components/Button.scss
+++ b/src/css/components/Button.scss
@@ -35,14 +35,13 @@ $Button-padding: 0 scaleGrid(2);
   font-family: $Button-font-family;
   font-size: $Button-font-size;
   font-weight: $Button-font-weight;
-  height: $Button-height;
+  min-height: $Button-height;
   line-height: $Button-line-height;
   margin: 0;
   padding: $Button-padding;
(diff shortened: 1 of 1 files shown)


bradcypert:
@saralohr are there any repercussions here that I should be aware of?

saralohr:
I think the issue is more "a lot of text is shoved in a small button" rather than the fixed height. From a design perspective buttons shouldn't have more than one line of text in them, and should be concise CTAs. It also presents issues when buttons are displayed side by side... they should all be the same height.

bradjteamsnap:
Hmm, I agree with Sara on the concise CTA preference. The `whitespace: nowrap` should take care of text breaking to the next line, and IMO buttons should always be set to `nowrap`. `min-height` can be a bit inconsistent between browsers I think, any alignment issues with button groups could probably be sorted out with flexbox properties on the parent element?

bradcypert:
That makes a lot of sense. The text that were wrangling with is '+ Copy to Team/Division'.

PR #638, opened 2021-05-19
Adding loader modal component

BillyFigueroa:
Adding a loader component

Screen Shot 2021-05-19 at 2 45 02 PM

This component is meant to help with indicating to a client that there is some process happening while they are waiting


src/css/components/Modal.scss
```diff
@@ -1,169 +1,40 @@
...
-  
+.Modal--closed {
+  display: none;
 }
 
-.Modal-bgDismiss {
-  height: 100%;
-  width: 100%;
-  position: absolute;
+.Modal__body {
```
bradcypert, line 30: I think the proper name for this would be `.Modal-body` based off of our CSS guide.

src/css/components/Modal.scss
```diff
@@ -1,169 +1,40 @@
...
 }
 
-@keyframes modalIn{
-  0% {
-    transform: scale(0.7);
-  }
-  100% {
-    transform: scale(1);
-  }
+.Modal__footer {
```
bradcypert, line 34: I think the proper name for this would be `.Modal-footer` based off of our CSS guide.

src/css/components/Modal.scss
```diff
@@ -1,169 +1,40 @@
...
 }
 
-@keyframes overlayOut{
-  0% {
-    opacity: 1;
-  }
-  100% {
-    opacity: 0;
-  }
+.Modal__header {
```
bradcypert, line 38: I think the proper name for this would be `.Modal-header` based off of our CSS guide.

bradcypert (commented):
I noticed the keyframes were removed from that SCSS file as well. Can we find a way to add those back in? Smooth transitions are 👌.

Also, I think I know why we have that modal scss file that is seemingly unused. Classic probably uses it. Classic uses the SCSS alone from TeamSnap UI but with V4, that pattern probably doesnt make sense anymore so its fine to remove those styles.

bradcypert (approved):

PR #652, opened 2021-05-26
Fix npm audit failures

ryanwi:
One high left, can't quite shake it

```
# Run  npm update y18n --depth 4  to resolve 1 vulnerability
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ High          │ Prototype Pollution                                          │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ y18n                                                         │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ gulp [dev]                                                   │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ gulp > gulp-cli > yargs > y18n                               │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://npmjs.com/advisories/1654                            │
└───────────────┴──────────────────────────────────────────────────────────────┘
```


bradjteamsnap (approved):

saralohr:
Since CSS Nano is deprecated we'll probably want to add a different minifier in it's place. Did you notice the difference in output size without it? I can pull it down and check it out too.

ryanwi:
@saralohr I did not check the size. a replacement sounds great if you have one?

saralohr:
Not off the top of my head, although on a quick search it appears css nano may still be valid packaged under postcss-cli. Not sure how that will work with sass though. I will look into the options. The min version is 114 kb, and the non-min is 161 kb, a 41% increase, so probably something we should do. But still not a huge payload.

ryanwi:
that is a relief it's not going to blow up the size

ryanwi:
does Netlify do any auto minify?

saralohr:
Good thinking! It does look like they provide that: https://docs.netlify.com/site-deploys/post-processing/

ryanwi:
I tried with the netlify.toml and in the ui (https://app.netlify.com/sites/ts-ui-assets/settings/deploys#asset-optimization) but doesn't seem to want to minify https://deploy-preview-652--ts-ui-assets.netlify.app/css/teamsnap-ui.css yet

saralohr:
Hrm, bummer. Are there different settings for deploy previews vs prod? Also it doesn't seem like I have access to TS Netlify account. Can I get added to that?

saralohr:
I just tried out `gulp-clean-css` locally with default options and it's down to 129kb. I'll see if I can configure it differently to get it any smaller, but otherwise I think this will work fine if Netlify doesn't work. I'll push up the change shortly.

saralohr:
I also noticed that we have a static dugout url set for our font assets.  I can't remember why we did this... perhaps for cacheing? But to have the same relative url for the font css, I think we'd need to move the css/themes files into the top level css directory so they can all use the same fonts/css partials with the same relative path. Trying to think if that would cause issues anywhere other than updating classic to use that path for SO theme. Looking at how we do theme switching in NJW.

saralohr (approved):

PR #669, opened 2021-06-04
Node version

saralohr:
Bacon and I ran into some issues being on different node and npm versions, so this locks those down with engines and nvmrc. Thanks @bacon for this!

The large diff on package-lock is for the regenerated file for the locked down versions.


l3aconator (approved):
LGTM, I hope 😂

ryanwi:
can you add 14 to the the matrix in the GH workflow too? https://github.com/teamsnap/teamsnap-ui/blob/master/.github/workflows/main.yml#L22

PR #670, opened 2021-06-04
Update node in matrix

saralohr:


saralohr:
Wasn't sure if 10 and 12 should also have been removed @ryanwi.

ryanwi:
I'll defer to the Triple B's on that one for final decision. Since we're no longer concerned about dugout, the 10 and 12 may not be needed?  Only need to be concerned about local builds and netlify at this point.

bradjteamsnap:
Is it not supposed to work on anything below 14 now? because the way I understand it is that github will build and run the tests on each of the node versions in that matrix. So if 10 and 12 would fail the it would just lead to failed CI tests if we leave those in, but the tests seem to be passing now (at least the ci ones)

saralohr:
It will probably work on versions below 14. We ran into an issue when Bacon was on 16.

bradjteamsnap:
Ah gotcha, I'd say leave them in bc it's not failing any tests. But could go either way. Looks like Netlify is on 14.7.0. BTW thanks for adding the nvmrc!

bradjteamsnap (approved):

saralohr:
Okay I'll leave them in for now. Bacon updated Netlify when it was failing the build with the version update. Team effort!

ryanwi:
before, when having to deploy with dugout, this was constrained because dugout was running node 8 or something. I believe it's no longer constrained, so as long as everything works well on 14, that seems like it's enough if it works there.

PR #673, opened 2021-06-07
Adding ability to display new flat-list division program structure

BillyFigueroa:
* Adding ability to view recipient list with the new flat list program/ seasons structure

Screen Shot 2021-06-07 at 11 28 51 AM

### DIFF
--- a/src/js/components/ToggleCheckboxList/ToggleCheckboxList.tsx
+++ b/src/js/components/ToggleCheckboxList/ToggleCheckboxList.tsx
@@ -25,6 +25,7 @@ type Props = {
   headerStatus: object;
   setActiveRows: React.Dispatch<React.SetStateAction<{}>>;
   setHeaderStatus: React.Dispatch<React.SetStateAction<{}>>;
+  sortOrder?: string;
 };
 
 const ToggleCheckboxList: React.FunctionComponent<Props> = ({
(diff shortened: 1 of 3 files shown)


src/js/components/ToggleCheckboxList/ToggleCheckboxList.tsx
```diff
@@ -25,6 +25,7 @@ type Props = {
   headerStatus: object;
   setActiveRows: React.Dispatch<React.SetStateAction<{}>>;
   setHeaderStatus: React.Dispatch<React.SetStateAction<{}>>;
+  sortOrder?: string;
```
bradcypert, line 28: thoughts on adding an enum to describe sort order? We could probably add that to the types folder in v4.

src/js/components/ToggleCheckboxList/ToggleCheckboxList.tsx
```diff
@@ -35,12 +36,31 @@ const ToggleCheckboxList: React.FunctionComponent<Props> = ({
...
+
+    if(sortOrder === 'desc') {
+      // Sort headers
+      list.sort((prev: ExpandableList, curr: ExpandableList) => prev.heading[0] < curr.heading[0] ? 1 : 0);
+      // Sort sub-rows
+      list.map(row => row.rows.sort((a, b) => a[0] < b[0] ? 1 : 0));
+    }
+  }
+
+  sort(list);
```
bradcypert, line 62: This should probably be in a use effect. Otherwise, this will sort on every render which is probably way overkill.

src/js/components/ToggleCheckboxList/ToggleCheckboxList.tsx
```diff
@@ -79,23 +100,24 @@ const ToggleCheckboxList: React.FunctionComponent<Props> = ({
...
   }
 
-  const toggleSubheadingRows = (heading: string, idx: number) => {
-    let rowData= [];
+  const toggleSubheadingRows = (heading: string, subheading: string, idx: number) => {
+    let rowData = [];
+    const headerName = `${heading}-${subheading}`;
 
-    if (heading in headerStatus && headerStatus[heading]['activeCount'] > 0) {
+    if (headerName in headerStatus && headerStatus[headerName]['activeCount'] >= 0) {
```
bradcypert, line 107: Optional, but I think you can shorten this to 
```suggestion
    if (headerStatus[headerName]?.['activeCount'] >= 0) {
```

bradcypert (approved):

PR #684, opened 2021-06-14
Adding archive option to component

BillyFigueroa:
Adding the archive option to the component to be able to display archived seasons

Screen Shot 2021-06-14 at 10 28 00 AM

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@teamsnap/teamsnap-ui",
-  "version": "4.0.0-alpha.40",
+  "version": "4.0.0-alpha.41",
   "description": "a CSS component library for TeamSnap",
   "main": "dist/js/index.js",
   "types": "dist/js/index.d.ts",
--- a/src/js/components/Tag/Tag.tsx
+++ b/src/js/components/Tag/Tag.tsx
@@ -1,11 +1,30 @@
 import * as React from 'react';
+import { Icon } from '../Icon';
 
 type Props = {
   text: string;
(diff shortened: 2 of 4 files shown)


bradcypert (commented):
What are your thoughts on finding a way to do this where it's more flexible (probably using some form of composition?). So far, this ToggleCheckboxList component has been very abstract, but now it's starting to become specific to seasons.

BillyFigueroa:
> What are your thoughts on finding a way to do this where it's more flexible (probably using some form of composition?). So far, this ToggleCheckboxList component has been very abstract, but now it's starting to become specific to seasons.

yea, I def felt a little dirty adding this. Do you have anything in mind? I think we can push this through and think about a solution to change this

bradcypert:
I'd really rather not push this through as is.  The `isArchived` flag doesn't really make sense outside of the scope of this one component use-case (so it's no longer generic) and TeamSnap UI is open source (and used by a few other people). It doesn't make sense for us to add a proprietary configuration for a component that's intended to be generic and for use everywhere.

It may be enough to pass in a list of tag components for additional tags.

```
 {
            "heading": "Micron Program 2",
            "subHeadings": [<Tag text=""2021 Summer Season"/>, <Tag icon="archived" text="Archived" />],
            "rows": []
}
```

What are your thoughts on this interface?

BillyFigueroa:
@bradcypert forgot this is open source. You are right.

I like an array or a function that returns something there. I ll look into both

bradcypert (approved):

PR #690, opened 2021-06-16
Clean up field component InputProps

### LINKED ISSUE #613: Field Component - InputProps are no longer used
With the refactoring to support checkboxes, it appears that the field component is no longer using inputProps. We should address this and remove these props if they're no longer required.

ocamacho-tsnap:
Clean up the InputProps in the Field Component

Closes #613


bradcypert (approved):
Thank you!

PR #693, opened 2021-06-17
minor fixes to text button and modal component

jondauz:
- Added `type=button` to close button of Modal component so that it does not submit form when closing modal. 
- Made modifications to `Button--text` utility class to make button look more like a text link.


jondauz:
I've also updated the Button component to accept a new type of "Link" which will display a text button.

bradcypert (approved):

PR #700, opened 2021-06-21
SO-4735-add-breadcrumbs-component

ocamacho-tsnap:
Add the Breadcrumbs component into v4

### DIFF
--- a/src/css/components/Breadcrumbs.scss
+++ b/src/css/components/Breadcrumbs.scss
@@ -2,13 +2,13 @@
 
 // * @define default Breadcrumbs
 
-.Nav-breadcrumb {
-  color: black;
-  font-size: 16px;
-}
+// Global default breadcrumbs colors
+$breadcrumb-primary-color: color-neutral(7) !default;
+$breadcrumb-secondary-color: color-neutral(10) !default;
 
-a {
-  color: black;
+.Nav-breadcrumb {
+  color: $breadcrumb-secondary-color;
+  font-size: scaleGrid(2);
 }
 
 .Breadcrumb {
@@ -17,17 +17,21 @@ a {
     margin-right: 10px;
     position: relative;
 
+    a {
+      color: $breadcrumb-secondary-color;
+    }
+
     &:last-child {
-      color: #5d7180;
+      color: $breadcrumb-primary-color;
     }
   }
 }
 
 .Icon-home {
-  color: #758693;
+  color: $breadcrumb-primary-color;
   margin-right: 2px !important;
 }
 
 .Icon-separator {
-  color: #768794;
+  color: $breadcrumb-primary-color;
 }
--- a/src/css/components/Breadcrumbs.scss
+++ b/src/css/components/Breadcrumbs.scss
@@ -1,0 +1,33 @@
+@import "../config/index.scss";
+
+// * @define default Breadcrumbs
+
+.Nav-breadcrumb {
+  color: black;
+  font-size: 16px;
+}
+
+a {
+  color: black;
+}
+
+.Breadcrumb {
+  li {
+    display: inline-block;
+    margin-right: 10px;
+    position: relative;
+
+    &:last-child {
+      color: #5d7180;
+    }
+  }
+}
+
+.Icon-home {
+  color: #758693;
+  margin-right: 2px !important;
+}
+
+.Icon-separator {
+  color: #768794;
+}
--- a/src/js/components/Breadcrumbs/Breadcrumbs.tsx
+++ b/src/js/components/Breadcrumbs/Breadcrumbs.tsx
@@ -1,0 +1,111 @@
+/**
+ * @name Breadcrumbs
+ *
+ * @description
+ *  A breadcrumbs component that will render the appropriate styles for breadcrumbs.
+ *
+ *
+ * @example
+ *  <Breadcrumbs breadcrumbs={breadcrumbs} home={home} />
+ *
+ *
+ */
+
+import * as React from "react";
+import * as PropTypes from "prop-types";
+import Icon from "../Icon/Icon";
+
+const propTypes = {
+  className: PropTypes.string,
+  breadcrumbs: PropTypes.array.isRequired,
+  home: PropTypes.any,
+  id: PropTypes.string,
+  style: PropTypes.object,
+};
+
+const Breadcrumbs: React.FunctionComponent<
+  PropTypes.InferProps<typeof propTypes>
+> = ({ className, breadcrumbs, home, id, style }) => {
+  const renderHome = () => {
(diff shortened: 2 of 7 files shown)


src/css/components/Breadcrumbs.scss
```diff
@@ -0,0 +1,33 @@
+@import "../config/index.scss";
+
+// * @define default Breadcrumbs
+
+.Nav-breadcrumb {
+  color: black;
```
bradcypert, line 6: try `color-neutral(10)` instead of black.

src/css/components/Breadcrumbs.scss
```diff
@@ -0,0 +1,33 @@
...
+  font-size: 16px;
```
bradcypert, line 7: There is a grid system in place that we can use here instead. I think this should be roughly `scaleGrid(2)` instead of `16px`.

src/css/components/Breadcrumbs.scss
```diff
@@ -0,0 +1,33 @@
...
+}
+
+a {
+  color: black;
+}
```
bradcypert, line 12: We cant add this style as this would make all anchors black across whatever app pulls this in. We either need to nest this specific style under the breadcrumb class, or find a different way to accomplish this.

src/css/components/Breadcrumbs.scss
```diff
@@ -0,0 +1,33 @@
...
+    position: relative;
+
+    &:last-child {
+      color: #5d7180;
+    }
+  }
+}
+
+.Icon-home {
+  color: #758693;
```
bradcypert, line 27: color-neutral(7);

src/css/components/Breadcrumbs.scss
```diff
@@ -0,0 +1,33 @@
...
+  margin-right: 2px !important;
+}
+
+.Icon-separator {
+  color: #768794;
```
bradcypert, line 32: I think (unless a designer has said otherwise) we want to this to be `color-neutral(7)` as well.

src/css/components/Breadcrumbs.scss
```diff
@@ -0,0 +1,33 @@
...
+}
+
+.Breadcrumb {
+  li {
+    display: inline-block;
+    margin-right: 10px;
+    position: relative;
+
+    &:last-child {
+      color: #5d7180;
```
bradcypert, line 21: I think this is also supposed to be `color-neutral(7)` but am not 100% positive.

src/js/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,111 @@
...
+> = ({ className, breadcrumbs, home, id, style }) => {
+  const renderHome = () => {
+    if (!home) {
+      return;
+    }
+
+    let content = (
+      <Icon
+        className="Icon Icon-home"
+        mods={null}
```
bradcypert, line 37: You can remove this prop since its just null.

src/js/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,111 @@
...
+        name="home"
+        otherProps={{}}
+        style={{}}
```
bradcypert, line 40: You can remove these prop since theyre just empty maps.

src/js/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+    return <li>{content}</li>;
+  };
+
+  const renderSeparator = () => {
+    return (
+      <li>
+        <Icon
+          className="Icon Icon-separator"
+          mods={null}
```
bradcypert, line 56: You can remove this since its set to null.

src/js/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,111 @@
...
+          name="right"
+          otherProps={{}}
+          style={{}}
```
bradcypert, line 59: You can remove these as well.

src/js/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,111 @@
...
+ */
+
+import * as React from "react";
+import * as PropTypes from "prop-types";
+import Icon from "../Icon/Icon";
+
+const propTypes = {
+  className: PropTypes.string,
+  breadcrumbs: PropTypes.array.isRequired,
+  home: PropTypes.any,
```
bradcypert, line 21: let's change this to capture the shape of the home object.

Something like:

```
PropTypes.shape({
  url: PropTypes.string
})
```

bradcypert (approved):

src/css/components/Breadcrumbs.scss
ocamacho-tsnap: Done!

src/css/components/Breadcrumbs.scss
ocamacho-tsnap: Done!

src/css/components/Breadcrumbs.scss
ocamacho-tsnap: Done!

src/css/components/Breadcrumbs.scss
ocamacho-tsnap: Done!

src/js/components/Breadcrumbs/Breadcrumbs.tsx
ocamacho-tsnap: Done!

src/js/components/Breadcrumbs/Breadcrumbs.tsx
ocamacho-tsnap: Done!

src/js/components/Breadcrumbs/Breadcrumbs.tsx
ocamacho-tsnap: Done!

src/css/components/Breadcrumbs.scss
ocamacho-tsnap: Done!

src/js/components/Breadcrumbs/Breadcrumbs.tsx
ocamacho-tsnap: Done!

src/js/components/Breadcrumbs/Breadcrumbs.tsx
ocamacho-tsnap: Done!

PR #733, opened 2021-07-06
SO-4735: Add ability to render children to Toolbar component

justin-prince:
Add ability to render children to Toolbar component and add story with breadcrumbs as Toolbar child component.

### DIFF
--- a/src/js/components/Toolbar/Toolbar.tsx
+++ b/src/js/components/Toolbar/Toolbar.tsx
@@ -61,100 +61,98 @@ const Toolbar: React.FunctionComponent<
   });
 
   return (
-    <>
-      <div className="Grid Toolbar">
-        {children && (
-          <div className="Grid-cell u-flexAlignSelfCenter u-sizeFit u-textNoWrap u-textEllipsis u-spaceRightLg">
-            {children}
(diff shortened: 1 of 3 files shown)


jondauz (approved):
Looks good

src/js/components/Toolbar/Toolbar.tsx
```diff
@@ -57,26 +62,34 @@ const Toolbar: React.FunctionComponent<
...
+            <Field
+              type="input"
+              formFieldProps={{
+                placeholder: "Find a program, season, or member",
+                leftIcon: <Icon className="Icon" name="search" />,
+              }}
+              name="Sample"
+            />
+        </div>
+      )}
```
bradcypert, line 82: We should probably remove this. The correct way to add a search after these changes would be to use component composition just like you did with the breadcrumbs example. Do you mind removing this piece and any associated bits (and bytes)?

justin-prince: @bradcypert, I've removed the search component from the Toolbars component and included it as a part of the Toolbar stories. Thanks.

bradcypert (approved):

PR #735, opened 2021-07-07
Minor fixes to Navbar found in staging

jondauz:
This PR is to fix bugs found in staging in regards to this ticket https://teamsnap.atlassian.net/browse/SO-4734.

- Added z-index for when overlay is open so that navigation is always on top.
- Removed hidden icon 
- Added max-width


justin-prince (approved):
Lookin' good.

PR #736, opened 2021-07-08
SO-4786: Fix spacing of paginated tables. 

jondauz:
Used `u-padBottomNone` to remove the padding applied from checkbox class. Opted to use utility class rather than using `isInline` prop due to some alignment issues within the table.

### DIFF
--- a/src/js/components/Table/Paginated/PaginatedTable.tsx
+++ b/src/js/components/Table/Paginated/PaginatedTable.tsx
@@ -147,6 +147,7 @@ const PaginatedTable: React.FunctionComponent<Props> = ({
                   }
                 },
               }}
+              mods="u-padBottomNone"
             />
           </div>
         ),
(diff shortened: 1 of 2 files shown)


src/js/components/Table/Paginated/PaginatedTable.tsx
```diff
@@ -147,6 +147,7 @@ const PaginatedTable: React.FunctionComponent<Props> = ({
                   }
                 },
               }}
+              mods="u-padBottomNone"
```
BillyFigueroa, line 150: Minor request, if you move this to line 138 below `name="select-all"` I think it would be easier to read

for example

```js
cols = [
      {
        name: "selected",
        label: (
          <div>
            <Checkbox
              name="select-all"
              mods="u-padBottomNone"
              inputProps={{
                checked: checkboxState,
                onClick: () => {
                  // if there is at least one row checked, clear all
                  if (checkboxState != CheckboxStates.FALSE) {
                    setSelected([]);
                  } else {
                    // select all rows if none are checked
                    setSelected(rows);
                  }
                },
              }}
            />
          </div>
        ),
        mods: "u-size1of12",
      },
      ...cols,
    ];
```

I will say it's optional for readability so I will just go ahead and approve this

BillyFigueroa (approved):

PR #743, opened 2021-07-08
change export method on icons with errors

bradjteamsnap:
These 3 icons are exported with `export default` while the rest are exported with `module.export`. We noticed that there are complier warnings/errors from these 3 icons from webpack in multiple projects that use teamsnap-ui.

Tested in local storybook to ensure icons are rendered correctly.


saralohr:
One of the build tasks was causing errors that this method fixed. I _think_ it was building storybook (vs running locally). Did you test with `yarn run build-storybook` too? The build task for storybook is different than the local task and had a bunch of errors initially. Also worth testing with `yarn run build` to make sure if you haven't already.

bradjteamsnap:
> One of the build tasks was causing errors that this method fixed. I _think_ it was building storybook (vs running locally). Did you test with `yarn run build-storybook` too? The build task for storybook is different than the local task and had a bunch of errors initially. Also worth testing with `yarn run build` to make sure if you haven't already.

Good call. I just ran `yarn build-storybook` and `yarn build` as well. There are warnings for `yarn build-storybook` but not different from those on master branch.

saralohr:
> There are warnings for yarn build-storybook but not different from those on master branch.

Okay. Honestly I went through so many hoops to get a clean build it's possible that the `export` was needed with one dependency set but not with a later one.

As long as the build is working, then 👍🏼

saralohr (approved):

saralohr:
So I should have looked back at the old PR rather than trusting my brain 😆 . I didn't actually add `export default` to those icons. I _did_ however add it to a bunch of the stories to get the new storybook working. One of which was the [Icons](https://github.com/teamsnap/teamsnap-ui/pull/719/files#diff-b7067bb02e2992399ecdd0839e49906e393e0ce4e4fd264c0e4b96e908e14f2dR8) story. But I don't understand why that would have any affect on these icons. So this may be an independent issue.

PR #752, opened 2021-07-14
SO-4734: More design fixes 

jondauz:
There was an issue with the Nav container when menu is opened with overlay the width would grow randomly. Also fixes some issues in classic where navigation wouldn't fix with text on smaller screens.


bradcypert (approved):
Please test these changes in Storybook with stakeholders so we can lock down the work here. We've got a lot of moving parts in TeamSnap UI at the moment and it'd be good for us to iron out working (as intended), releasable versions of these components 😄

PR #754, opened 2021-07-14
So 4735 toolbars with breadcrumbs

justin-prince:
Make the Breadcrumbs component more flexible by allowing users to simply pass in an array of links, text, or whatever and just render those out as breadcrumbs.


bradcypert (approved):
This looks great! Thank you so much for looking into this!

Issue #691, opened 2021-06-16
Our TypeScript code needs a linter


bradcypert:
We have a fair amount of inconsistency between fundamental syntax choices with TeamSnap UI. Ideally, we can setup a linter configuration and address the inconsistencies.

bradcypert:
@BillyFigueroa Any suggestions on a linter configuration for a baseline? I know this is an area that you're passionate about, so I'd love for you to lead this one if you'd like to :)

BillyFigueroa:
Still working on this

PR #762, opened 2021-07-21
Add linter

BillyFigueroa:
Adding linter for v4 so that our code base can have a more uniform coding style

This addresses and fixes issue #691


jondauz (approved):
Looks good to me.

bradjteamsnap (approved):
Nice!

BillyFigueroa:
> Nice! 

Going to be adding a git pre hook with husky soon maybe in a diff pull request.

bradcypert (approved):
Can we add a section in the README under linters about our use of eslint?

bradcypert:
@BillyFigueroa worth mentioning that precommit hook will be clientside only and I dont think it gets committed with the repo. I was experimenting with this a while back to prevent direct commits to specific branches with another repo and ended up adding a folder called hooks with the precommit hook there, and then adding a makefile that set up the hooks for people. That being said, people have to opt into it, which is fine, but I wouldn't assume everyone will use it.

On that note, I probably won't. I'm a regularly committer (hence my favor of squashing commits 😃). If I have to wait for a  lint everytime I commit, that'll likely slow me down tremendously, haha. I will still adhere to the linter, ultimately though!

BillyFigueroa:
> @BillyFigueroa worth mentioning that precommit hook will be clientside only and I dont think it gets committed with the repo. I was experimenting with this a while back to prevent direct commits to specific branches with another repo and ended up adding a folder called hooks with the precommit hook there, and then adding a makefile that set up the hooks for people. That being said, people have to opt into it, which is fine, but I wouldn't assume everyone will use it.
> 
> On that note, I probably won't. I'm a regularly committer (hence my favor of squashing commits 😃). If I have to wait for a lint everytime I commit, that'll likely slow me down tremendously, haha. I will still adhere to the linter, ultimately though!

Thanks for that insight Brad. What do you recommend then we do for people not using VSCode?

BillyFigueroa:
> Can we add a section in the README under linters about our use of eslint?

What exactly should I add regarding use? that we are using airbnb etc? that sort of stuff?

BillyFigueroa:
@bradcypert I added a way to run prettier command line by using `lint:format` and then added some more information in the `README.MD` and made some quick updates to `yarn` commands because, unlike `NPM` we don't need

bradcypert (approved):

PR #800, opened 2021-08-06
So 4925 birthday filter

LeonardoHabitzreuter:
Created a date variant for PaginatedTable filters
Now user can filter by a range of dates or selecting specific years...


bradcypert (commented):
Also whats the deal with that "No Date" option? I dont think we want that, right?

BillyFigueroa:
> 
> 
> Also whats the story of that "No Date" option? I dont think we want that, right?

The force (eslint/ prettier) is coming

jondauz (approved):

PR #817, opened 2021-08-20
SO-5013: Fix show admin conditional rendering

jondauz:
https://teamsnap.atlassian.net/browse/SO-5013

### DIFF
--- a/src/js/components/Toolbar/Toolbar.tsx
+++ b/src/js/components/Toolbar/Toolbar.tsx
@@ -87,14 +87,16 @@ const Toolbar = ({
             <div className="u-posAbsolute Toolbar-flyout">{accountBody}</div>
           )}
         </div>
-        <div>
-          <Button type="link" style={btnStyle} onClick={() => setActiveMenu(Menu.Admin)}>
-            {showAdmin && <Icon className="Icon u-spaceLeftLg" name="wrench" style={iconStyle} />}
-          </Button>
-          {activeMenu === Menu.Admin && (
-            <div className="u-posAbsolute Toolbar-flyout">{adminBody}</div>
-          )}
-        </div>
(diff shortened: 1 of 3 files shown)


src/js/components/Toolbar/Toolbar.tsx
```diff
@@ -87,14 +87,16 @@ const Toolbar = ({
...
         </div>
-        <div>
-          <Button type="link" style={btnStyle} onClick={() => setActiveMenu(Menu.Admin)}>
-            {showAdmin && <Icon className="Icon u-spaceLeftLg" name="wrench" style={iconStyle} />}
-          </Button>
-          {activeMenu === Menu.Admin && (
-            <div className="u-posAbsolute Toolbar-flyout">{adminBody}</div>
-          )}
-        </div>
+        {showAdmin && (
```
BillyFigueroa, line 90: Might need to check if it is a ghost inspector user also?

jondauz: How do I check for ghost inspector user?

BillyFigueroa: @jondauz hmmm never mind, this is something you can check in teamsnap spas

BillyFigueroa (approved):

src/js/components/Toolbar/Toolbar.tsx
bradcypert: Yeah, as this is open source and available to people outside if TeamSnap, we don't want to check for anything specific to what we use at TeamSnap. If a conditional needs to be available to a component where we need data like that, it needs to be available via a prop, which it is with the showAdmin prop. In fact, the implementation we have currently might even be too specific, but we can worry about that another time.

PR #846, opened 2021-09-13
Add support for setting combobox values from props

bradcypert:
The combobox component did not have a way to set its state from a prop, so I've added that. Additionally, I've wired up the paginated table to read the value from the context for a given field to do just that :)

### DIFF
--- a/src/js/components/ComboBox/Combobox.tsx
+++ b/src/js/components/ComboBox/Combobox.tsx
@@ -37,6 +37,9 @@ const propTypes = {
       label: PropTypes.string,
     })
   ),
+  selected: PropTypes.arrayOf(
+    PropTypes.any
+  ),
   onChange: PropTypes.func,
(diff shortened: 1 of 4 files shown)


src/js/components/ComboBox/Combobox.tsx
```diff
@@ -94,30 +98,42 @@ const ComboBox = ({
...
+        toggleFlyout(false);
+        onChange(selectedFilters);
+      }
     } else {
       clearFilters();
     }
   };
 
+  React.useEffect(() => {
+    if (!!selected && (selected.length !== 0) && selected.some((item) => !(selectedFilters || []).includes(item))) {
```
LeonardoHabitzreuter, line 128: Is it possible to create a named function that returns this result? I think could be a good idea to make this piece of code more understandable

LeonardoHabitzreuter (approved):

PR #867, opened 2021-09-20
Fix is issue with totalItems re-render

jondauz:
Screen Shot 2021-09-20 at 9 23 41 AM

In TeamSnap SPAs in the `loadData` function when we call `setTotalItems(members.length);` on line 68 the PaginatedTable re-renders again with an empty filter list which is incorrect. In order to fix that issue I'm thinking it may be a better idea to set totally items within the PaginatedTable component itself since in the `loadData` function returns a list of items, in the PaginatedTable we could set `totalItems` from the length of that list. 

I've also made a change to the "Filter" toggle button to conditionally add `u-hidden` to the filters section instead of the button rendering the filters and combobox when filters are toggle to be shown. This will make it so that the filters persist when the filters shown or hidden.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@teamsnap/teamsnap-ui",
-  "version": "4.0.0-alpha.67",
+  "version": "4.0.0-alpha.68",
   "description": "a CSS component library for TeamSnap",
   "main": "dist/js/index.js",
   "types": "dist/js/index.d.ts",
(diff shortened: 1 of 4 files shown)


bradcypert:
The load data function will not necessarily return the list of all items. If we're using "virtual" pagination, that's true, but if we're actually paginating on the server, I dont think that will work.

bradcypert:
We should sync up on the changes here and decide which of our 3 PRs we want to take changes from

PR #872, opened 2021-09-22
Checkbox label html

jondauz:
Give the ability to access html as a label for combo box component.

### DIFF
--- a/src/js/components/InputControl/InputControl.tsx
+++ b/src/js/components/InputControl/InputControl.tsx
@@ -69,9 +69,12 @@ const InputControl = ({
         {...inputProps}
         checked={value} // has to come after spreading input props to support indeterminate
       />
-      <label className="Checkbox-label" htmlFor={name} {...labelProps}>
-        {label}
-      </label>
+      <label
+        dangerouslySetInnerHTML={{ __html: label as string }}
+        className="Checkbox-label"
+        htmlFor={name}
+        {...labelProps}
+      />
     </div>
   );
 };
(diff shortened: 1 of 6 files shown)


src/js/components/InputControl/InputControl.tsx
```diff
@@ -69,9 +69,12 @@ const InputControl = ({
         {...inputProps}
         checked={value} // has to come after spreading input props to support indeterminate
       />
-      <label className="Checkbox-label" htmlFor={name} {...labelProps}>
-        {label}
-      </label>
+      <label
+        dangerouslySetInnerHTML={{ __html: label as string }}
```
bradcypert, line 73: Why not just allow the label to be either a string or a react node?

jondauz: @bradcypert I've refactored the code so that we aren't using `dangerouslySetInnerHTML`

bradcypert (changes requested):
Please hold off on this. Ill create a PR against it to propose some changes.

bradcypert:
Hey @jondauz! What are your thoughts on this instead? https://github.com/teamsnap/teamsnap-ui/pull/877 Im not 100% convinced, but I do like the idea of the simpler filter structure. This may be less generic though, as if you pass a JSX node to the filter, you will only get the label of any string nodes in the JSX.

I think I understand why the subtext was added (we might want to rename that though). I am good with either approach :)

bradcypert (approved):

PR #883, opened 2021-09-27
Remove padding and max-height from fullscreen modal.

justin-prince:


src/css/components/Modal.scss
```diff
@@ -34,12 +34,19 @@
 
   .Modal-content {
     position: relative;
-    width: auto;
-    top: auto;
-    left: auto;
-    border-radius: none;
+    width: 100vw;
+    height: 100vh;
+    max-height: none;
```
BillyFigueroa, line 39: Might this bring the same issue as `border-radius`? maybe use `0` instead?

BillyFigueroa (approved):

PR #887, opened 2021-09-28
Filters and DateFilter fixes

jondauz:
- https://teamsnap.atlassian.net/browse/SO-4925 - Date filter fixes
- https://teamsnap.atlassian.net/browse/SO-5210 - One filter should be open at a time
- https://teamsnap.atlassian.net/browse/SO-5214 - "Clear" should clear out text box in combo box as well.

### DIFF
--- a/src/js/components/ComboBox/Combobox.tsx
+++ b/src/js/components/ComboBox/Combobox.tsx
@@ -26,6 +26,7 @@ import { Button } from '../Button';
 import { Checkbox } from '../Checkbox';
 import { Field } from '../Field';
 import { Icon } from '../Icon';
+import { isEqual } from 'lodash';
 
 const propTypes = {
   name: PropTypes.string.isRequired,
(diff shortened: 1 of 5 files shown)


src/js/components/ComboBox/Combobox.tsx
```diff
@@ -26,6 +26,7 @@ import { Button } from '../Button';
 import { Checkbox } from '../Checkbox';
 import { Field } from '../Field';
 import { Icon } from '../Icon';
+import { isEqual } from 'lodash';
```
bradcypert, line 29: Apparently we already have lodash as a dependency for TeamSnap UI, but I dont think depending on that is a good idea ultimately. We'll probably want to remove this at some point, but its likely way beyond the scope of this PR.

src/js/components/Table/Paginated/PaginatedTable.stories.tsx
```diff
@@ -396,7 +396,7 @@ function mapData(item) {
     birthdate: item.birthdate ? item.birthdate.toLocaleDateString('en-US') : 'undefined',
     age: `${item.age}`,
     programs: item.activePrograms.map((p, idx) => <div key={idx}>{p.name}</div>),
-    team: item.team.name,
+    team: item?.team?.name,
```
bradcypert, line 399:
```suggestion
    team: item.team?.name,
```

We dont need the first null check operator. I know it's just a `?` but theres a little bit of overhead added in the compiled JS everytime we use that, so using it unnecessarily is something we want to avoid :)

src/js/components/Table/Paginated/PaginatedTable.tsx
```diff
@@ -17,6 +17,7 @@ import { assert } from '../../../utils/assert';
 import { Button } from '../../Button';
 import { Panel } from '../../Panel';
 import DateFilter from './DateFilter';
+import { isEqual } from 'lodash';
```
bradcypert, line 20: We're not using this in this file, right? Let's remove it if not.

bradcypert (approved):

src/js/components/ComboBox/Combobox.tsx
jondauz: Absolute agree with you. Might be a good thing to visit during cooldown. I think for now in order to keep pushing through, if okay, I'm going to opt to keep as is.

PR #896, opened 2021-09-30
SO 5231 create multi select card component

analauhou:
https://user-images.githubusercontent.com/85572162/135907993-7beb06b3-3ea2-45c0-b9af-3ac6c2a2a5e3.mov

Reference of old changes:

https://user-images.githubusercontent.com/85572162/135534210-88c7f837-722b-4973-a343-d3674bda1a9c.mov

### DIFF
--- a/src/css/components/SelectCard.scss
+++ b/src/css/components/SelectCard.scss
@@ -6,6 +6,7 @@
     color: color-neutral(9);
     border: 1px solid color-neutral(4);
     font-size: $tu-base-fontSize;
+    cursor: pointer;
 
     &:hover {
         border: 1px solid color-primary(6);
--- a/src/css/components/SelectCard.scss
+++ b/src/css/components/SelectCard.scss
@@ -1,30 +1,40 @@
 .Card {
     border-radius: 5px;
-    width: 15%;
-    height: 25%;
+    width: 100%;
+    height: auto;
     text-align: center;
     color: color-neutral(9);
     border: 1px solid color-neutral(4);
     font-size: $tu-base-fontSize;
 
     &:hover {
-        box-shadow: 0 8px 16px 0 color-neutral(4);
+        border: 1px solid color-primary(6);
+    }
+
+    &.is-selected {
+        border: 1px solid color-primary(6);
+        .Card--icon {
+            color: color-primary(6);
+        }
+    }
+
+    input {
+        display: inline-block;
+        position: absolute;
+        width: 100%;
+        height: 100%;
+        left: 0px;
+        top: 0px;
+        cursor: pointer;
+        opacity: 0 !important;
     }
 }
 
 .Card--container {
     padding: 20px 20px;
+    position: relative;
 }
 
 .Card--icon {
     color: color-neutral(8);
-    margin-bottom: 10px;
 }
-
-.Card--selected {
-    border: 1px solid color-primary(6);
-    .Card--icon {
-        color: color-primary(6);
-    }
-}
-
--- a/src/css/components/SelectCard.scss
+++ b/src/css/components/SelectCard.scss
@@ -1,0 +1,30 @@
+.Card {
+    border-radius: 5px;
+    width: 15%;
+    height: 25%;
+    text-align: center;
+    color: color-neutral(9);
+    border: 1px solid color-neutral(4);
+    font-size: $tu-base-fontSize;
+
+    &:hover {
+        box-shadow: 0 8px 16px 0 color-neutral(4);
+    }
+}
+
+.Card--container {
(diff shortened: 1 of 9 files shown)


BillyFigueroa:
Hey Ana,

GREAT WORK! I love the video and how it looks

I would like to recommend a change. Passing children as a prop the way you are doing can get complicated. I think rather than doing something like this...

```JavaScript
const content = (
  <div>
    <p>Hello!</p>
    <p>Select me!</p>
  </div>
)

return (
  <SelectCard children={content} style={{padding: '20px'}}/>
)
```

we should build it so that you can do this...

```JavaScript
<SelectCard style={{padding: '20px'}}>
  <div>
    <p>Hello!</p>
    <p>Select me!</p>
  </div>
</SelectCard>
```

What do you think? The other question is, does the UI call for any animations on this? You might want to ask @andrewmeacham

src/css/components/SelectCard.scss
```diff
@@ -0,0 +1,30 @@
...
+    color: color-neutral(9);
+    border: 1px solid color-neutral(4);
+    font-size: $tu-base-fontSize;
+
+    &:hover {
+        box-shadow: 0 8px 16px 0 color-neutral(4);
+    }
+}
+
+.Card--container {
```
saralohr, line 15: I haven't been keeping up with CSS PRs lately, (and it's not noted anywhere easy to find!) but we should be using [SUIT CSS naming conventions](https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md#componentname).

Example: 

`ComponentName` - primary class
`ComponentName-descendantElement` - descendant elements
`ComponentName--modifierName` - styles for modifiying main component
`Component.is-stateName` - state class

So in this case these should be descendents not modifiers: `Card-container`, `Card-icon` etc.

And in the `Card--selected` case, I think this should be a [state class](https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md#componentnameis-stateofcomponent) rather than a modifier.

So the class would be `.is-selected` and for specificity it would be styled as `Card.is-selected`.

Thank you and please reach out if you have any questions!

src/js/components/SelectCard/SelectCard.tsx
```diff
@@ -0,0 +1,65 @@
...
+  };
+
+  const cardClasses = getClassName(
+    className,
+    selected && `Card--selected`,
+    mods
+  );
+
+  return (
+    <div className={cardClasses} style={style} {...props} onClick={handleClick}>
```
saralohr, line 39: I wonder about writing this as an `input type=checkbox` with the checkbox hidden vs a `div` to support keyboard interaction natively and to use semantic HTML. Thoughts?

src/css/components/SelectCard.scss
BillyFigueroa: @saralohr how long ago was the decision to use this made and was anyone who is currently working here who was a part of the decision still around?

Would like to know why this was chosen over BEM instead

src/css/components/SelectCard.scss
```diff
@@ -0,0 +1,30 @@
+.Card {
+    border-radius: 5px;
+    width: 15%;
```
saralohr, line 3: We may want to set this to `width: 100%` and `height: auto` for maximum flexibility. Then use a wrapping element like `Grid` or just [`u-flex` utilities](https://teamsnap-ui-patterns.netlify.app/patterns/utilities/flex.html) to quickly compose elements on the page in different layouts.

andrewmeacham:
Looks great, @analauhou! Can you add the `cursor: pointer`  👆 to the hover state?

Also, can we remove the check icon swap for this version? This will give us more flexibility in the future. Sorry for the miscommunication on my end!

Screen Shot 2021-10-01 at 12 14 57 PM

https://www.figma.com/file/r9jD0MuXnMYO9edeLnuOtj/360-Registration?node-id=181%3A2433

src/css/components/SelectCard.scss
saralohr: This is an alternate syntax for BEM, but still subscribes to the Block-Element-Modifier ideology, and all the CSS in TeamSnap UI _should_ follow it. I made the decision when we built V1 of TeamSnap UI.

saralohr: Now that we are relying more on React components the CSS class naming is less important, but still nice to be consistent if possible.

analauhou:
@saralohr @andrewmeacham Thanks for the feedback, please take a look again when you have a chance.

BillyFigueroa (approved):

saralohr (approved):

PR #900, opened 2021-10-01
So 5213

LeonardoHabitzreuter:
Ticket:
https://teamsnap.atlassian.net/browse/SO-5213

The purpose of this PR is to improve the feedback for users when PaginatadTable has no data to show

### DIFF
--- a/src/js/components/Table/Paginated/PaginatedTable.tsx
+++ b/src/js/components/Table/Paginated/PaginatedTable.tsx
@@ -63,6 +63,7 @@ const propTypes = {
   rowsAreSelectable: PropTypes.bool,
   searchPlaceholder: PropTypes.string,
   defaultSort: PropTypes.string,
+  placeholder: PropTypes.string,
 };
 
 const Filter = (
@@ -127,6 +128,7 @@ const PaginatedTable: PaginatedTableProps = ({
   searchPlaceholder,
   paginationPlacement,
   defaultSort,
+  placeholder,
 }) => {
   assert(
     !(filters.length && paginationPlacement === Placement.Top),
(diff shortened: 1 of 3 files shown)


src/js/components/Table/Paginated/PaginatedTable.tsx
```diff
@@ -173,7 +175,7 @@ const PaginatedTable: PaginatedTableProps = ({
       setIsLoading(false);
       if (data) setDataSet(data);
 
-      if (currentTotalItems) setTotalItems(currentTotalItems);
+      if (typeof currentTotalItems === 'number') setTotalItems(currentTotalItems);
```
bradcypert, line 178: Why is this change needed?

LeonardoHabitzreuter: We need to update the totalItems state even when the "currentTotalItems" is 0, before this change the if condition was returning false

LeonardoHabitzreuter: I thought on code setTotalItems(currentTotalItems || 0) but I don't know if this can be some side effects

src/js/components/Table/Paginated/PaginatedTable.tsx
```diff
@@ -63,6 +63,7 @@ const propTypes = {
   rowsAreSelectable: PropTypes.bool,
   searchPlaceholder: PropTypes.string,
   defaultSort: PropTypes.string,
+  placeholder: PropTypes.string,
```
jondauz, line 66: I would suggest to change the param name to be something more descriptive than `placeholder` such as `noResultsText`.

jondauz (changes requested):

bradcypert (approved):

PR #915, opened 2021-10-07
Make it so that text buttons have a transparent background.

jondauz:
Before:
Screen Shot 2021-10-07 at 2 22 45 PM

After: 
Screen Shot 2021-10-07 at 2 22 54 PM


bradcypert (approved):
Oof. Sorry about that. I didnt even realize there was a white background there.

PR #940, opened 2021-10-15
ci(github actions): add support for Semantic Release for Alpha builds

bradcypert:
Additionally, upgraded to Yarn 3 as part of getting support for this working.


.releaserc
```diff
@@ -0,0 +1,6 @@
+{
+  branches: [
+    {name: 'master'},
```
ttacon, line 3: If we don't want to run the release process on this branch [as per here](https://github.com/teamsnap/teamsnap-ui/pull/940/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R5), should we remove this line?

package.json
```diff
@@ -71,6 +71,7 @@
     "babel-register": "^6.26.0",
     "browser-sync": "^2.26.14",
     "css-loader": "^5.2.6",
+    "cz-conventional-changelog": "^3.3.0",
```
ttacon, line 74: ❤️

.releaserc
bradcypert: You are totally correct. good catch!

PR #944, opened 2021-10-19
Add testId props to components to explicitly call it out as an optional yet supported prop

justin-prince:
Before this addition, data-testid props could be add via the otherProps prop in most components. This change hopes to further encourage its usage by making it explicitly clear that it's an option. Hopefully, consumers of the components will use it to ensure more robust testing.


bradcypert:
This is a great idea!

LeonardoHabitzreuter:
> @justin-prince not sure why thats failing. I was able to build it successfully locally.

On my machine the lint command is returning 10 errors

bradcypert (approved):
Thanks for adding this Justin. Even if we dont add tests directly into this repository, providing those test ids to other consumes seems like the courteous thing to do.

PR #966, opened 2021-10-28
chore(storybook): organize storybook in atomic design, add MDX support

bradcypert:
Not merging this one without a discussion regarding if we want to continue down this path ;)

Please check out the deploy preview and see if this is something we want to continue with!

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -1,6 +1,6 @@
 module.exports = {
   stories: [
-    '../src/**/*.stories.@(tsx|js)',
+    '../src/**/*.stories.@(tsx|js|mdx)',
   ],
   addons: [
     '@storybook/addon-actions',
--- a/src/js/components/Avatar/Avatar.stories.tsx
+++ b/src/js/components/Avatar/Avatar.stories.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import Avatar from './Avatar';
 
 export default {
-  title: 'Avatar',
+  title: 'Design System/Atoms/Avatar',
 };
 
 export const Default = () => (
--- a/src/js/components/Avatar/Avatar.stories.tsx
+++ b/src/js/components/Avatar/Avatar.stories.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import Avatar from './Avatar';
 
 export default {
-  title: 'Design System/Atoms/Avatar',
+  title: 'Components/Data Display/Avatar',
 };
 
 export const Default = () => (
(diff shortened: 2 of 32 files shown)


bradcypert:
@saralohr Part of my goal with adding support for MDX stories was to be able to inclde that type of information. For example, the Colors story is an MDX file and, while there is no React component specifically for colors, it still is technically an atom. 

I'd love to get more thoughts on how we could structure things if not atomically.

saralohr:
That's great we have MDX support! I totally missed the colors story under atoms. Looks sweet. I think that is a good example though of how design documentation can get lost if it's mixed in with components. I'd vote for having React components under a "Components" directory, maybe with some further sub-directories underneath it, and have design docs under another directory. For the moment, if colors is the only design doc we have, maybe it lives at the top level. 

I would also like to add CSS documentation for things like the utility classes that won't ever have a component associated with them. This also feels like something that would live independently from components for easier differentiation.

bradjteamsnap:
Yeah I agree with Sara about separating the patterns & utils from the Components. But not against the atom/molecule/organism structure in theory. In practice I've seen it get out of control and not stay sorted well. Can we at least add something to the pull request template with clarification on how to sort (ie: "I have made sure my changes follow the agreed upon atomic taxonomy out lined here [link]")?

bradcypert:
What are your thoughts around categories like:
- Forms
- Data Display
- Feedback
- Surfaces
- Navigation

instead of the atomic levels?

bradcypert:
@saralohr my original vision here was to make it so you could just identify whether it was an atom, molecule, or organism but I like that suggestion a lot.

What about two top level folders?
- Patterns
- Components

And specifically for components, for the time being, we can organize them around the following subcategories:
- Forms
- Data Display
- Feedback
- Surfaces
- Navigation

bradjteamsnap:
I always like the way our docs were organized in Baseweb (Uber), though it had it's own problems.
https://baseweb.design/

jondauz:
> @saralohr my original vision here was to make it so you could just identify whether it was an atom, molecule, or organism but I like that suggestion a lot.
> 
> What about two top level folders?
> 
> * Patterns
> * Components
> 
> And specifically for components, for the time being, we can organize them around the following subcategories:
> 
> * Forms
> * Data Display
> * Feedback
> * Surfaces
> * Navigation

I like this much better than the "atomic" approach. The approach above is more straight forward and easy to understand in my opinion.

BillyFigueroa:
I like the ideas suggested already. When dealing with patterns, sometimes its difficult for a person to identify things if they are not really familiar with the pattern. As seen, we were already having a hard time trying to figure out what is at atom vs molecule etc sticking to just more normal naming conventions is probably a good idea

saralohr (approved):
Looks great!

saralohr (approved):
Looks great!

bradjteamsnap (approved):
yeah this makes much more sense to me. Would love to add a section around utility classes in the future!

jondauz (approved):

PR #972, opened 2021-11-03
Add id props for form input components

justin-prince:
Currently, Checkbox and Radio components use the name property as the HTML ID attribute. It's common
for those elements to share a name but have unique IDs. This commit adds that functionality.

### DIFF
--- a/src/js/components/CheckboxPill/CheckboxPill.tsx
+++ b/src/js/components/CheckboxPill/CheckboxPill.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import { Grid } from '../Grid';
 import { Cell } from '../Cell';
-import { Pill } from '../Pill';
+import { Pill, PillStatus } from '../Pill';
 import { Field } from '../Field';
 
 const CheckboxPill = () => {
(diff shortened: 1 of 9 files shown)


analauhou (approved):

justin-prince:
I created this PR so that `name` and `id` props could be independent of one another. In some cases, the `id` HTML attribute was being set to the `name` prop. That could cause issues when several form inputs share the same name yet are expected to use unique IDs. This is not uncommon with groups of radio or checkbox inputs. Duplicate IDs within the same document are invalid but, more importantly, cause issues when a `label`'s `for` attribute references an ID shared by two input elements.

This change maintains BC by only setting the `id` attribute to the `id` prop _if_ it is passed into the component. Otherwise, it continues to use the `name` prop as the `id` attribute on the input element.

Another solution is to forward the `inputProps` prop downward through the Checkbox and Radio components where an `id` property could be added (à la `inputProps={{ id: "favorite-color--red }}`). With this approach, the `id` attribute on the InputControl component is overridden and somehow de-duped before it gets rendered out as HTML.

Any thoughts? Any gotchas I'm not considering?

src/js/components/CheckboxPill/CheckboxPill.tsx
```diff
@@ -14,7 +14,7 @@ const CheckboxPill = () => {
   return (
     <Grid>
       <Cell>
-        <Pill align="center">
+        <Pill align="center" status={checked ? PillStatus.ACTIVE : null}>
```
justin-prince, line 17: This was added just to make the tests pass. :)

jondauz (approved):

PR #975, opened 2021-11-09
Rap 135: Add a drawer component

justin-prince:


BillyFigueroa:
@justin-prince I think the component needs to be added to the exported list if we want to use it. Not sure I see that happening

src/js/components/Drawer/Drawer.stories.tsx
```diff
@@ -0,0 +1,356 @@
...
+
+      <Pill>Drawer is {isDrawerOpen ? 'Open' : 'Closed'}</Pill>
+
+      <Drawer
+        placement={Placement.Bottom}
+        open={isDrawerOpen}
+        closeFn={closeDrawer}
+        testId={`Drawer-storybook-default`}
+      >
+        <Panel mods={`u-spaceMd`}>
```
BillyFigueroa, line 226: This doesn't need braces

src/css/components/Modal.scss
```diff
@@ -49,12 +49,6 @@
   }
 }
 
-// Prevent the body from scrolling while the modal is open.
-body.Modal--open {
-  overflow: hidden;
-  height: 100%;
-}
-
```
justin-prince, line 57: Moved this to a dedicated, more general `.scroll-lock` class.

src/js/hooks/useOnClickOutside.ts
```diff
@@ -0,0 +1,21 @@
...
+    document.addEventListener('touchstart', listener);
+
+    return () => {
+      document.removeEventListener('mouseup', listener);
+      document.removeEventListener('touchstart', listener);
+    };
+  }, [ref, handler]);
+};
+
+export { useOnClickOutside };
```
justin-prince, line 21: Given a ref to a DOM element, this hook will fire a provided callback handler whenever a click is trigger from outside of the element.

src/js/hooks/useBodyScrollLock.ts
```diff
@@ -0,0 +1,11 @@
...
+
+const useBodyScrollLock = (isInitiallyLocked = false, scrollLockClass = 'scroll-lock') => {
+  const [isBodyScrollLocked, setIsBodyScrollLocked] = React.useState(isInitiallyLocked);
+
+  document.body.classList.toggle(scrollLockClass, isBodyScrollLocked);
+
+  return { isBodyScrollLocked, setIsBodyScrollLocked };
+};
+
+export { useBodyScrollLock };
```
justin-prince, line 11: This hook is a lean wrapper around adding a scroll lock class to the DOM body, returning the current state and the means to update that state. Useful for improved UX with modal dialogs and this new drawer component.

src/css/components/Drawer.scss
```diff
@@ -0,0 +1,98 @@
+$drawer-box-shadow-offset: 6px;
+$drawer-box-shadow-blur: 12px;
+$drawer-box-shadow-color: rgba(0, 0, 0, 0.2);
+$drawer-box-shadow-blur-color: #{$drawer-box-shadow-blur} #{$drawer-box-shadow-color};
+$drawer-border: 1px solid #ddd;
+
+.Drawer {
+  background-color: #fff;
```
bradcypert, line 8: We'll probably want to pull this out to variable for themeing purposes. Actually, it would probably be good find common colors for surfaces and pull them out as well. Might be beyond the scope of this, though.

src/js/components/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,84 @@
...
+  }, [open, allowBodyScroll]);
+
+  const showDrawerClass = open ? 'Drawer--open' : '';
+  const placementAxisClass = [Placement.Left, Placement.Right].includes(placement)
+    ? 'Drawer--placement-x'
+    : 'Drawer--placement-y';
+
+  const classes = getClassName(
+    'Drawer',
+    `Drawer--placement-${placement}`,
```
bradcypert, line 62: I'm not sure how feasible this is, but it may help with readability if you group your useEffects together and your const declarations together if possible.

bradcypert (approved):

src/css/components/Drawer.scss
justin-prince: Totally agreed. We might want to consider updating our Sass to use `@use` over `@import` at some point in the near future. I'm not sure when they plan on dropping `@import` completely. That might be a good time to tighten up things like this.

src/js/components/Drawer/Drawer.tsx
justin-prince: Done. Thanks!

PR #982, opened 2021-11-11
Modal updates

saralohr:
- Add customClose option for modal
- Fullscreen layout styles and example in storybook
- Modify modal layout component so content scrolls under heading and scroll bar doesn't break layout

**Before:**
modal-current

**After:**
modal-updates


src/js/components/Modal/Modal.stories.tsx
```diff
@@ -161,101 +162,140 @@ export const Fullscreen = () => {
...
-          <i>eget</i> molestie dui. Praesent blandit dolor. Sed non quam. In vel <b>lacinia</b> mi
-          sit amet augue congue elementum. Morbi in ipsum sit amet pede facilisis laoreet. Donec
-          lacus <b>augue</b> nunc, <b>sit</b> viverra nec.
-        </p>
+        <div className="u-padLg"
+          style={{
+            flex: '1',
+            overflow: 'auto'
+          }}
+        >
```
saralohr, line 178: This is the scrollable part of the modal content in the wizard.

src/js/components/Modal/Modal.stories.tsx
```diff
@@ -161,101 +162,140 @@ export const Fullscreen = () => {
...
+            <i>eget</i> molestie dui. Praesent blandit dolor. Sed non quam. In vel <b>lacinia</b> mi
+            sit amet augue congue elementum. Morbi in ipsum sit amet pede facilisis laoreet. Donec
+            lacus <b>augue</b> nunc, <b>sit</b> viverra nec.
+          </p>
+        </div>
+        <div
+          style={ {
+            flex: '0 0 100px',
+          } }
+        >
```
saralohr, line 270: This is the footer container. I think ideally this would be broken out into a `Modal-footer` class as part of the main modal component, but this is a way forward within the current structure being used for the wizard.

src/css/components/Modal.scss
```diff
@@ -22,18 +22,28 @@
   min-width: 50%;
   height: auto;
   max-height: 90vh;
-  overflow: scroll;
+  overflow: hidden;
```
saralohr, line 25: This will force the modal-body to scroll instead of the main modal container, allowing the header to be fixed and the scroll bar to be inset, preserving the rounded corner appearance.

saralohr:
@BillyFigueroa @justin-prince still not having luck getting org-frontend running locally, but these are essentially the changes I wanted to apply there.

src/js/components/Modal/Modal.tsx
```diff
@@ -59,6 +61,7 @@ const Modal: React.FC<Props> = ({
           <div className="u-sizeFill">
             <h2 data-testid="modal-heading">{heading}</h2>
           </div>
+          {customClose}
```
saralohr, line 64: Not sure if this is the best way to handle this. Basically just want a way to add a custom button in place of the showClose default "x"

BillyFigueroa: @saralohr what is the problem with not displaying the default close button and passing some elements to the header the way I did in staging?

If we have a way to provide a close button by passing it to the header (which accepts components and jsx), why do we need to add another option?

saralohr: The header element is an `h2` nested in a layout container already providing the structure for a close button via `showClose`. So if we pass to the `header` the markup gets pretty weird forcing us to nest a whole new layout container inside of the `h2`.

```
<div class="Modal-header u-flex u-flexJustifyBetween">
    <div class="u-sizeFill">
          <h2 data-testid="modal-heading">
           // this is where `header` populates
           // this is the content we'd have to pas into the h2 to have a header and a custom close button
            <div class="u-flex u-flexJustifyBetween u-flexAlignItemsCenter>
               <div>Header Content</div>
              <button>Save and Close</button>
           </div>
         </h2>
      </div>
      // this is where default close button lives and achieves the layout we want already
</div>
```

BillyFigueroa: I prefer the flexibility to have a dev create a layout rather in the header component than updating the modal with another close option @bradcypert can you comment on this please

bradcypert: I can definitely see both preferences here, but I think I tend to lean towards Sara's implementation. Keeping the close icon separate from the "Header" makes a lot of sense to me, and it's probably way more accessible markup (you're not "closing the header" you're "closing the modal"). We'll want to keep the generated markup from TeamSnap UI as semantic as possible.

src/js/components/Modal/Modal.stories.tsx
```diff
@@ -161,101 +162,140 @@ export const Fullscreen = () => {
   return (
     <>
       <Button onClick={() => setIsOpen(true)}>Test Modal</Button>
-
       <Modal
         heading="Full bleed"
+        customClose={<Button size='large'>Save and Exit</Button>}
```
bradcypert, line 167: thoughts on just making this property `closeButton` instead of `customClose`?

saralohr: Sure, that makes sense!

src/js/components/Modal/Modal.tsx
BillyFigueroa: @bradcypert thanks for the input.

src/js/components/Modal/Modal.stories.tsx
BillyFigueroa: @saralohr let me know if you are going to make the change or if you want me to make it

saralohr: Thanks, I think I can make it @BillyFigueroa though I may hit you up with some questions.

src/js/components/Modal/Modal.tsx
```diff
@@ -58,8 +60,8 @@ const Modal: React.FC<Props> = ({
         <div className="Modal-header u-flex u-flexJustifyBetween">
           <div className="u-sizeFill">
             <h2 data-testid="modal-heading">{heading}</h2>
-          </div>
-          {showClose && (
+          </div>          
+          {showClose && closeButton ? closeButton : (
```
saralohr, line 64: Changed the approach here to link `showClose` and `closeButton` to create three states:
- `showClose` is false: no close icon
- `showClose` is true and `closeButton` is false:  uses default "x" icon
- `showClose` and `closeButton` are true:  uses custom close Button

I thought this worked better than making `showClose` and `closeButton` independent, since they shouldn't both be true.

@BillyFigueroa in the case of the wizard, Drew and I like having the "x" on the first three steps for an easy cancel (no save). So in the wizard code when `closeButton` is `null` the default `showClose` icon will show up.

Any concerns with this approach @justin-prince @BillyFigueroa @bradcypert ?

BillyFigueroa: @saralohr I like the 3 states

src/js/icons/task-alt.ts
```diff
@@ -0,0 +1 @@
+module.exports = {"metadata":{"xmlns":"http://www.w3.org/2000/svg","viewBox":"0 0 24 24","name":"task-alt"},"source":"<path fill=\"none\" d=\"M0 0h24v24H0z\"/><path d=\"M21.29 5.89l-10 10a.996.996 0 01-1.41 0l-2.83-2.83a.996.996 0 111.41-1.41l2.12 2.12 9.29-9.29a.996.996 0 011.41 0c.4.39.4 1.02.01 1.41zm-5.52-3.15c-1.69-.69-3.61-.93-5.61-.57-4.07.73-7.32 4.01-8.01 8.08a10.009 10.009 0 0011.19 11.66c3.96-.51 7.28-3.46 8.32-7.31.4-1.47.44-2.89.21-4.22-.13-.8-1.12-1.11-1.7-.54-.23.23-.33.57-.27.89.22 1.33.12 2.75-.52 4.26-1.16 2.71-3.68 4.7-6.61 4.97-5.1.47-9.33-3.85-8.7-8.98.43-3.54 3.28-6.42 6.81-6.91 1.73-.24 3.37.09 4.77.81a1.003 1.003 0 00.93-1.78c-.27-.12-.54-.25-.81-.36z\"/>"}
```
BillyFigueroa, line 1: add a newline at the end here

saralohr: This is a generated file and I see none of the others have a new line... seems like it's okay without?

justin-prince (approved):

PR #992, opened 2021-11-17
fix(drawer component): fix issue with propTypes in the Drawer component

justin-prince:
Fix mismatch between defaultProps and propTypes. Attach propTypes to Drawer component.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@teamsnap/teamsnap-ui",
-  "version": "4.0.0-alpha.95",
+  "version": "4.0.0-alpha.96",
   "description": "a CSS component library for TeamSnap",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@teamsnap/teamsnap-ui",
-  "version": "4.0.0-alpha.95",
```
bradcypert, line 3: While you're free to do this, this does not need to be updated anymore. It's processed from the git tag by semantic release before publishing. If we want to have it automatically update the version, thats something I can look at, but manually updating this is no longer needed :)

bradcypert (approved):

package.json
justin-prince: That's terrific news. Thanks!

PR #1008, opened 2021-11-26
Load table trigger

LeonardoHabitzreuter:
A way to reload the Table data when apollo cache is updated


bradcypert (commented):
This is a really good idea! My only hesitancy is on the name `loadDataDep`. Maybe if we added some JSDoc comments around that property and indicated that if you use that prop and it changes, we'll refresh the data for you.

andresaguilarbaires (approved):
Besides Brad's comment, LGTM.

PR #1089, opened 2022-01-19
SO-5525: New Popup tooltip component 

jondauz:
Screen Shot 2022-01-19 at 12 19 29 PM

### DIFF
--- a/src/js/components/Popup/PopupTooltip.tsx
+++ b/src/js/components/Popup/PopupTooltip.tsx
@@ -3,17 +3,19 @@ import * as PropTypes from 'prop-types';
 
 const propTypes = {
   text: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
-  popupStyle: PropTypes.object,
   testId: PropTypes.string,
+  ariaDescribeBy: PropTypes.string,
 };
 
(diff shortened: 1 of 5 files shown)


src/js/components/Popup/PopupTooltip.tsx
```diff
@@ -0,0 +1,56 @@
...
+  popupStyle: PropTypes.object,
+  testId: PropTypes.string,
+};
+
+type Props = React.FunctionComponent<PropTypes.InferProps<typeof propTypes>>;
+
+const PopupTooltip: Props = ({ text, testId, children }) => {
+  const buttonRef = React.useRef<HTMLButtonElement | null>(null);
+  const tooltipRef = React.useRef<HTMLDivElement | null>(null);
+  const [pos, setPos] = React.useState<{ top: number; left: number }>({ top: 0, left: 0 });
```
LeonardoHabitzreuter, line 15: I don't think we need to set the type in this "pos" state, it seems redundant for me.
Another thing, could you change it to "position"? It's more readable than just "pos"

LeonardoHabitzreuter:
Greate work, @jondauz !

LeonardoHabitzreuter (approved):

PR #1098, opened 2022-01-24
fix(modal component): nested ternary

justin-prince:
Remove a nested ternary in JSX


saralohr:
Thanks Justin this is much cleaner!

saralohr (approved):

PR #1103, opened 2022-01-26
fix linting errors

jondauz:
Current v4's auto publishing to NPM library is broken. The assumption is that it is broken because there was various linting errors which prevented the library from auto publishing.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -8,7 +8,7 @@
     "es6": true
   },
   "parserOptions": {
-    "project": "./tsconfig.json"
+    "project": "./tsconfig.eslint.json"
   },
   "rules": {
     "react/require-default-props": "off",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@teamsnap/teamsnap-ui",
-  "version": "4.0.0-alpha.110",
```
bradcypert, line 3: Its fine to update this, but it shouldnt be necessary if the auto-deployment is working as expected.

bradcypert (approved):
These changes look good. Thanks for fixing these errors. I'll keep an eye on the CI process and see if its still failing to deploy.

package.json
BillyFigueroa: @jondauz weren't you and trey looking at the auto process not working? What came out of that

BillyFigueroa (approved):

saralohr (approved):
Thanks for fixing this up!!

PR #1106, opened 2022-01-27
fix(package.json): correct version

jondauz:
trying to get the correct version going again so that we can have semantic releases again


wnoonan (approved):
(•_•)
( •_•)>⌐■-■
(⌐■_■)

PR #1107, opened 2022-01-27
ci(lint-pull-request.yml): add new github workflow

jondauz:
Add a new github action workflow that can be used to ensure PR title follows semantic release specs"

### DIFF
--- a/.github/workflows/lint-pull-request.yml
+++ b/.github/workflows/lint-pull-request.yml
@@ -1,0 +1,17 @@
+name: 'Lint PR'
+
+on:
+  pull_request_target:
+    types:
+      - opened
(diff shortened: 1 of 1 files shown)


.github/workflows/lint-pull-request.yml
```diff
@@ -0,0 +1,17 @@
+name: 'Lint PR'
+
+on:
+  pull_request_target:
```
bradcypert, line 4: Not sure if there's a way to scope this down to just v4, but with push targets you can do something like this to specify branches: https://github.com/PyreStudios/steward/blob/main/.github/workflows/deploy_docs.yml#L5

PR #1120, opened 2022-02-02
fix(checkbox): Update size to use pixel dimensions

saralohr:
I recently updated the checkbox design and the percent positioning of the checkbox must not have been hitting a solid pixel value because it was moving around a bit dependent on browser size.

This updates to calculate size and position in pixels based on the overall size of the checkbox so positioning should be consistently correct.

Before:
Screen Shot 2022-02-02 at 12 00 08 PM

After:
Screen Shot 2022-02-02 at 11 54 23 AM
Screen Shot 2022-02-02 at 11 54 31 AM
Screen Shot 2022-02-02 at 11 54 43 AM

### DIFF
--- a/src/css/components/Checkbox.scss
+++ b/src/css/components/Checkbox.scss
@@ -41,6 +41,9 @@ $Checkbox-bg-color: $cu-foreground !default;
     &:before {
       height: $size;
       width: $size;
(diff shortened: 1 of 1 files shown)


justin-prince (approved):
Looks good!

PR #1152, opened 2022-03-07
feat(component): adding modify classes to component

BillyFigueroa:
Adding the ability to modify the component through classes being passed to this new attribute


justin-prince (approved):
Looking good!

PR #1179, opened 2022-03-28
Fix ToggleCheckboxList issue with whitespace

dbhaasis:
If item used for sub list had trailing spaces. It would not select show those items as selected if they were displayed. The reason is that innerText would always be trimmed if open but not if the item was not visible.


src/js/components/ToggleCheckboxList/ToggleCheckboxList.tsx
```diff
@@ -163,20 +163,21 @@ const ToggleCheckboxList: React.FunctionComponent<Props> = ({
       },
     };
 
+    console.log('set active rows row click', newActiveList);
```
jondauz, line 166: Is console.log needed?

jondauz (approved):

PR #1190, opened 2022-04-04
feat(paginatedtable): extended core features or paginated table

jondauz:


src/js/components/Table/Paginated/PaginatedTable.tsx
```diff
@@ -451,30 +464,55 @@ const PaginatedTable: PaginatedTableProps = ({
...
+          </>
+        ) : (
+          <FilterContext.Provider value={{ activeFilters, setActiveFilters }}>
+            <div className="u-size7of8">
+              {filters.map((Item, index) => (
+                <Item key={index} isLast={index === filters.length - 1} />
+              ))}
+            </div>
+            <div className="u-size1of8 u-textRight u-spaceRightMd">
+              <Button
```
LeonardoHabitzreuter, line 499: What do you think about wrapping this code into another func? If you do that, you won't need to duplicate those lines... It would be like this:
```
<FilterContext.Provider value={{ activeFilters, setActiveFilters }}>
    {getFiltersAndButtons()}
</FilterContext.Provider>
```

LeonardoHabitzreuter (approved):

PR #1228, opened 2022-04-27
fix(combox and paginatedtable): add ability to provide link for tooltip

jondauz:
Make icon for tooltip clickable.

### DIFF
--- a/src/js/components/ComboBox/Combobox.tsx
+++ b/src/js/components/ComboBox/Combobox.tsx
@@ -49,6 +49,7 @@ const propTypes = {
   otherProps: PropTypes.object,
   disabled: PropTypes.bool,
   tooltip: PropTypes.element,
+  tooltipLink: PropTypes.string,
 };
 interface Filter {
   value: string;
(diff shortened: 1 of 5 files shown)


dbhaasis (approved):
Only question is if clicking on the icon rather than hover is supposed to follow link or not. I think it will follow link. Also wondering about how this works for mobile touchscreens.

jondauz:
@dbhaasis On touch devices when clicking the icon it will open the link in a new page, if a link is provide. The tooltip itself will open in both cases.

src/js/components/ComboBox/Combobox.tsx
```diff
@@ -58,6 +59,23 @@ interface Filter {
 
 type Props = PropTypes.InferProps<typeof propTypes>;
 
+const TooltipWrapper = ({ tooltipLink, children }) => {
+  if (!tooltipLink) {
+    return <>{children}</>;
+  }
+
+  return (
+    <a href={tooltipLink} target="_blank" rel="noreferrer">
```
wnoonan, line 68: was looking up noreferrer, do you think noopener would be beneficial here too? https://www.reliablesoft.net/noreferrer-noopener/#noopener

wnoonan (commented):
just had a question in a comment but otherwise lgtm!

PR #1341, opened 2022-08-25
feat(component): re-adding attributes

BillyFigueroa:
Re-adding some attributes for `ReactDatePicker`


src/js/components/DatePicker/DateTimePicker.tsx
```diff
@@ -108,7 +118,15 @@ const DateTimePicker = ({ containerClasses, datetime, inputClasses, ...rest }: P
 
 DateTimePicker.defaultProps = {
   containerClasses: '',
+  dateFormat: '',
```
analauhou, line 121: do we want to add `datetime` here for defaultprops too?

analauhou (approved):

PR #1378, opened 2022-09-28
Fix Panel to stop ignoring id prop passed through table

LeonardoHabitzreuter:


jondauz (approved):

LeonardoHabitzreuter:
feat(panel): Fix Panel to stop ignoring id prop passed through table

PR #1391, opened 2022-10-10
Popup refactoring

LeonardoHabitzreuter:
* Using Functional component instead of class
* Using Typescript instead of prop-types
* Handling hover state using CSS instead of JS to avoid recreating components when the user hovers over popup actions

### DIFF
--- a/src/js/components/Popup/Popup.stories.tsx
+++ b/src/js/components/Popup/Popup.stories.tsx
@@ -31,7 +31,7 @@ export const PopupActions = () => (
       <PopupAction
         text={<Icon name="more-horiz" />}
         actions={actions}
-        direction={['down', 'center']}
+        direction={['down', 'overlay']}
         popupStyle={{ width: '150px' }}
       />
     </div>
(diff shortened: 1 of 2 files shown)


analauhou (approved):
👍

src/js/components/Popup/Popup.stories.tsx
```diff
@@ -31,7 +31,7 @@ export const PopupActions = () => (
       <PopupAction
         text={<Icon name="more-horiz" />}
         actions={actions}
-        direction={['down', 'center']}
+        direction={['down', 'overlay']}
```
jondauz, line 34: Does the change from `center` to `overlay` introduce some regression issues? I think it would be a good idea to just make sure that anywhere that uses `PopupAction` doesn't use  `{['down', 'center']}. If so be sure to have a PR ready to make those changes before pushing v4

src/js/components/Popup/Popup.stories.tsx
```diff
@@ -44,7 +44,7 @@ export const PopupActionsNoHover = () => (
...
```
jondauz, line 47: Same as above.

jondauz (approved):
Approved but please take a look at my comment before publishing.  Nice work!

src/js/components/Popup/Popup.stories.tsx
LeonardoHabitzreuter: I changed from `center` to `overlay` cause there is no `center` option... Maybe someone remove it and forgot to update the stories

PR #1459, opened 2023-03-13
feat: Fixed ExpandableGroup B2B-320

analauhou:


CatalanoWebDevelopment (changes requested):
This definitely looks good! However, I try to keep logic out of the render if possible. So, in this instance, I would drop the `if/else` conditionals into the actual `onLabelChange()` and `onLabelBlur()` methods. In the render, ideally we just see `onBlur={() => onLabelBlur()}` for example

CatalanoWebDevelopment (approved):
Once you drop the logic into the respective `onBlur` and `onChange` handlers, this bad boy should be good to go!

analauhou:
> Once you drop the logic into the respective `onBlur` and `onChange` handlers, this bad boy should be good to go!

cool thanks @CatalanoWebDevelopment I made this change

PR #1615, opened 2023-07-20
B2B-1408 Closed Helper Text on Expandable Groups

CatalanoWebDevelopment:
Add closed helper text to `ExpandableGroup`.

Screenshot 2023-07-20 at 11 18 42 AM

### DIFF
--- a/src/js/components/ExpandableGroup/ExpandableGroup.tsx
+++ b/src/js/components/ExpandableGroup/ExpandableGroup.tsx
@@ -17,7 +17,7 @@ const propTypes = {
   onLabelBlur: PropTypes.func,
   children: PropTypes.node.isRequired,
   mods: PropTypes.string,
-  closedHelperText: PropTypes.string || PropTypes.node,
+  groupHelperText: PropTypes.string || PropTypes.node,
 };
 
 type Props = PropTypes.InferProps<typeof propTypes>;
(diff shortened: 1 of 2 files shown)


src/js/components/ExpandableGroup/ExpandableGroup.tsx
```diff
@@ -110,6 +112,10 @@ const ExpandableGroup = ({
           </div>
         </div>
 
+        {!expanded && closedHelperText && (
+          <div className="u-size4of12 u-flex u-flexJustifyEnd">{closedHelperText}</div>
```
BillyFigueroa, line 116: This should have some shade of grey for the text. Regarding what shade, maybe the mock has the shade or you can ask Isa

src/js/components/ExpandableGroup/ExpandableGroup.tsx
```diff
@@ -17,6 +17,7 @@ const propTypes = {
   onLabelBlur: PropTypes.func,
   children: PropTypes.node.isRequired,
   mods: PropTypes.string,
+  closedHelperText: PropTypes.string || PropTypes.node,
```
BillyFigueroa, line 20: It might be a better name to call this `groupHelperText` because if design later decides we want this text to display when the panel is open `closedHelperText` doesn't make sense there

PR #1735, opened 2024-01-10
fix otherProps error in the console

LeonardoHabitzreuter:
Screen Shot 2024-01-10 at 13 06 18


jondauz (approved):
Thank you for fixing this. Its very much appreciated

